PR #13, opened 2019-08-07
Setup JPA configuration.

GuN41K:
Do not forget create DB in your PC (name of DB you can check in application.properties, in JDBC link).


nazarvladyka:
It's better to use - "createDatabaseIfNotExist=true" in "spring.datasource.url"
E.g. spring.datasource.url=jdbc:mysql://localhost:3306/GreenCity?createDatabaseIfNotExist=true&useSSL=false&serverTimezone=UTC

So each user won't need to worry about creating a new database

PR #15, opened 2019-08-07
#14 create main packages

GuN41K:


tarasgl:
The application.properties file should not be loaded to the remote repository.
This config file contains config info(database url, password, etc.) that is unique for each developer

tarasgl:
resolve conflict before merge

Issue #34, opened 2019-08-14
Create UserRepo, UserOwnSecurityRepo.


GuN41K:
I need to create this repositories for cooperation with DB.
Parent task: #29 
[brach](https://github.com/SoftServe-Social-Projects/GreenCity/tree/%2334-Create-UserRepo-UserOwnSecurityRepo)

Issue #21, opened 2019-08-09
Propose new cafe


ghost:
As a registered user I want to propose a new café to be added to the map/list, so that information about cafes is up-to-date and full. 

**Mockups:**

**Acceptance Criteria:**
1. GIVEN a registered user is successfully logged in to the system, WHEN he clicks on "Add place" button at top right corner of the screen, THEN system shall open a modal window with the following:
a. input field "Type"
b. input field "Name"
c. input field "Address"
d. input field "Working times"
e. Buttons "Add" and "Cancel"

2. Provided user clicks on field "Type", system shall open a drop-down with the following elements:
a. Cafe
b. Restaurant
c. Add new

3. Provided user selects "Add new" option in the Type field, system shall let him enter a new type of place.

3. Fields "Name" and "Address" shall be char + digit fields. User can input max number of 30 chars into each field.

4. Provided user clicks on "Working times" field, system shall open a pop-up, where user has to choose opening and closure times. There shall be buttons "OK" and "Cancel".

5. All fields are mandatory to be populated. Provided at least one field is not populated, then "Add" button shall be disabled.

6. Provided user clicks on "Add" button, system shall close the modal window, add the proposed place to the system and display it in the "Proposed places" tab of the Admin page.

7. Provided user clicks on "Cancel" button, system shall close the modal window and not save data entered in the fields (if any).

**Assumptions and Constraints:**

**TBD:**

rsyvulko:
Not relevant.

PR #37, opened 2019-08-14
#34 create user repo, user own security repo

GuN41K:
#34


src/main/java/greencity/entities/Category.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @Column(nullable = false, unique = true, length = 100)
+    private String name;
+
+    @ManyToOne
+    private Category category;
```
tarasgl: Category class has a category field?
Why?

src/main/java/greencity/entities/Category.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @OneToMany(mappedBy = "category")
+    private List<Category> categories = new ArrayList<>();
```
tarasgl: List of categories in Category class?
Some sort of nextlevel pattern :)

src/main/java/greencity/entities/OpeningHours.java
```diff
@@ -0,0 +1,37 @@
...
+
+import javax.persistence.*;
+import java.time.LocalTime;
+
+@Entity
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class OpeningHours {
```
tarasgl: Why OpeningHours is a distinct entity?
Maybe it is better to push these fields to the Place entity? (Just thinking, not sure if you need that)
Or do you want to save here opening time for each day of the week for every place? (Then it makes sense)

src/main/java/greencity/entities/Location.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.NoArgsConstructor;
+
+import javax.persistence.*;
+
+@Entity
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class Location {
```
tarasgl: I think it is better to not have Location a separate table in the database for location but either put latitude and longitude fields to the Place entity (Useless OneToOne relation). These fields will be used almost always so maybe better remove this relation?

src/main/java/greencity/entities/Category.java
GuN41K: This is named Recursion Reference. One category can has many sub-category: Food-category has cafe-category and more. And cafe-category can has many sub-categories.

src/main/java/greencity/entities/Category.java
GuN41K: Same. Is Recursion reference.

src/main/java/greencity/entities/OpeningHours.java
GuN41K: "Or do you want to save here opening time for each day of the week for every place? (Then it makes sense)" - yes. Each day can has different opening time.

src/main/java/greencity/entities/Category.java
tarasgl: First of all - you don't need categories to have such a tree-like structure. In business logic of your app there is only need for a main category (such as come-with-own-cup, e-charge, etc) and some of these categories can have (or cannot) sub-categories (refering to the #21 issue come-with-own-cup can have type(sub-category) such as cafe, restaurant, coffee-shop etc) and that is it. You never should go deeper in this hierarchy of types in your app, so Recursion reference is an overkill for your app and it will be more complicated to work with in the future(read acceptance criteria for #21 issue that you need to choose type from dropdown menu)

So you can make two tables Category and Sub-Category with OneToMany relation.

OR if you want to save the possibility to have unlimited depth tree structure, at least name this field parentCategory, as you did in the Comment entity
But again, in Comment Recursion reference is obvious to use and mathes business logic. Here it is an overkill

src/main/java/greencity/dto/user_own_security/UserRegisterDto.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.validation.constraints.Email;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.Pattern;
+
+@Data
+@RequiredArgsConstructor
+public class UserRegisterDto {
+
+    @NotBlank
+    @Length(max = 20)
```
Jarvizzik: Maybe, it would be nice to add also min length and exception message.

src/main/java/greencity/entities/Category.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+
+@Entity
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class Category {
```
Jarvizzik: Add @EqualsAndHashCode and exclude mutable fields.

src/main/java/greencity/dto/user_own_security/UserRegisterDto.java
```diff
@@ -0,0 +1,30 @@
+package greencity.dto.user_own_security;
+
+import lombok.Data;
+import lombok.RequiredArgsConstructor;
+import org.hibernate.validator.constraints.Length;
+
+import javax.validation.constraints.Email;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.Pattern;
+
```
diru4ova: @  Data already consists @RequiredArgsConstructor
You probaly meant @NoArgsConstructor  or @AllArgsConstructor?
https://projectlombok.org/features/Data

src/main/java/greencity/entities/Category.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import javax.persistence.*;
+import java.util.ArrayList;
+import java.util.List;
+
+
```
diru4ova: Comment regarding all the Entities:
Are ALL the fields in ALL Entities not changable? 
If any value can change - it must be excluded from hashcode and equals method. 
Please consider something like that: @EqualsAndHashCode(exclude = {"changableField1", "changableField2", ...})

src/main/java/greencity/entities/OpeningHours.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @Column(nullable = false)
+    @DateTimeFormat(pattern = "HH:mm")
+    private LocalTime openTime;
+
+    @Column(nullable = false)
```
diru4ova: Be careful with this formatting. 
It can throw ConversionFailedException during serialisation or deserialisation. Or both.
If this is just for nice date presentation - you can change it at front-end. 
If this have wider purpose - check this very well. Or think about some Converter

Issue #35, opened 2019-08-14
Create UserService, UserOwnSecurityService and UserOwnSecurityController


GuN41K:
I have to create this services for our business logic.
Parent task: #29 
[branch](https://github.com/SoftServe-Social-Projects/GreenCity/tree/%2335-Create-UserService-UserOwnSecurityService-UserOwnSecurityController)

PR #47, opened 2019-08-18
#35 Create user service, user own security service, user own security controller

GuN41K:


Zap1999:
Are you sure it should be master branch?

src/main/java/greencity/controller/UserOwnSecurityController.java
```diff
@@ -0,0 +1,24 @@
...
+@AllArgsConstructor
+public class UserOwnSecurityController {
+
+    private UserOwnSecurityService service;
+
+    @PostMapping
+    public void register(@Valid @RequestBody UserRegisterDto dto) {
+        service.register(dto);
+    }
+}
```
Zap1999: I think it would be better to have a return here. For example, to return a status or a validation error.

GuN41K: if we have some errors we throw exception that will be catch by error handler. So if no any errors that mean that user successfully registered. That mean that i need not return anything.

Zap1999: https://medium.com/hashmapinc/rest-good-practices-for-api-design-881439796dc9

Issue #44, opened 2019-08-16
Create submit email logic while registration.


GuN41K:
We have to register our user. User has to write email. We have to submit this email. So i have to create this logic that sent some submit-email to user email.

Parent task: #29 
[branch](https://github.com/SoftServe-Social-Projects/GreenCity/tree/%2344-Create-submit-email-logic-while-registration)

PR #48, opened 2019-08-18
#44 Create submit email logic while registration

GuN41K:


GuN41K:
if you want to check this functionality you have to write your email and password in application.properties

src/main/java/greencity/service/impl/UserOwnSecurityServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.stereotype.Service;
+
+@Service
+@AllArgsConstructor
+public class UserOwnSecurityServiceImpl implements UserOwnSecurityService {
+
+    private UserOwnSecurityRepo repo;
+    private UserService userService;
+    private VerifyEmailService verifyEmailService;
+
```
diru4ova: In my opinion DTO should be used at Controller layer. Service should receive pure Entity or data. 
But up tp U

Issue #38, opened 2019-08-15
Add block, make admin, deactivate registered user


RostyslavKhasanov:
Parent task #27 
Branch - [ #38-Add-block-remove-make-admin-registered-users](https://github.com/SoftServe-Social-Projects/GreenCity/tree/%2338-Add-block-remove-make-admin-registered-users)

Should be add: 

- UserController, UserService.
- Method findAll which return list of users by page. Sort list by email in alphabetical order (A-Z).
- Method for block user.
- Method for update role.
- Method for remove user.
- Test for UserServiceImpl methods

PR #56, opened 2019-08-20
#38 add block, remove, make admin registered users

RostyslavKhasanov:


src/test/java/greencity/services/impl/UserServiceImplTest.java
```diff
@@ -0,0 +1,76 @@
...
+                User.builder()
+                        .firstName("test")
+                        .lastName("test")
+                        .email("test@gmail.com")
+                        .role(ROLE.USER_ROLE)
+                        .isBlocked(false)
+                        .lastVisit(LocalDateTime.now())
+                        .dateOfRegistration(LocalDateTime.now())
+                        .build();
+        userService.save(user);
```
tarasgl: Better use mockito instead of saving test user in the database.
When you are writing tests for service layer, you need to mock all the other classes (such as repository), so you are sure every time when this test fails it is only because of code in service class (but not because of db failture and so on)

RostyslavKhasanov: resolved

Issue #46, opened 2019-08-17
Approve/decline proposed cafes


nazarvladyka:
Releted to #23 
Include #42 
Create method and controller which will change status of place.
branch - [#46-approve-decline-cafes_backend](https://github.com/SoftServe-Social-Projects/GreenCity/tree/%2346-approve-decline-cafes_backend)

PR #58, opened 2019-08-21
#46 approve decline cafes backend

nazarvladyka:


src/main/java/greencity/service/impl/PlaceServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    private final PlaceRepo placeRepo;
+
+    /**
+     * Get current date and time in zone.
+     *
+     * @param zoneId - zone id.
+     * @return LocalDateTime object.
+     * @author Nazar Vladyka.
+     */
+    public static LocalDateTime getDateTime(String zoneId) {
```
tarasgl: Maybe move this method to a DateTimeService because it does nothing related to Place

nazarvladyka: done

Issue #43, opened 2019-08-16
Add list of cafes near the map (Entitys,Sevices,Controllers)


MarianMilian:
Add repository, service and test services

PR #75, opened 2019-08-24
#43 add list of cafes near the map

MarianMilian:


src/main/java/greencity/dto/openingHours/OpeningHoursDto.java
```diff
@@ -0,0 +1,34 @@
...
+    private LocalTime openTime;
+
+    @NotNull
+    @JsonFormat(pattern = "HH:mm", shape = JsonFormat.Shape.STRING)
+    @ApiModelProperty(dataType = "java.lang.String")
+    private LocalTime closeTime;
+
+    @NotNull private WeekDay weekDay;
+
+    public OpeningHoursDto(OpeningHours openingHours) {
```
tarasgl: I think that you use modelmapper to map entity to dto and vice versa

MarianMilian: Yes , you are right , but it isn't my part of my code, I just resolved merge conflict in this class

Issue #68, opened 2019-08-23
Sign-in the application


GuN41K:
I have to create logic that allow you to sign in user.
Parent: #66 
[branch](https://github.com/SoftServe-Social-Projects/GreenCity/tree/%2368-Sign-in-the-application)

PR #80, opened 2019-08-25
#68 sign in the application

GuN41K:


src/main/java/greencity/service/impl/UserOwnSecurityServiceImpl.java
```diff
@@ -29,41 +33,51 @@
     @Transactional
     @Override
     public void register(UserRegisterDto dto) {
+        log.info("begin");
```
tarasgl: In the future provide more informative messages))))

Issue #26, opened 2019-08-11
Access to cafe detail information


nhrab:
As an unregistered user I want to be able to access to the detail information about a place on the map so that I can choose the place to go.

**Mockups:**
![image](https://user-images.githubusercontent.com/53810790/62904792-63cee100-bd70-11e9-990d-f7e0b2df3f8a.png)

**Acceptance criteria:**
1. WHEN Unregistered or Registered user clicks on the one of test places, THEN he/she should be able to see small pop-up with detailed info about place (test data):
a. Name
b. Address (City, Street, Number of building)
c. Working times (day; hh:mm-hh:mm)
d. Discount
e. Feedback
f. Unactive star-icon in order to add this place to Favorites (as Registered user)

**Assumptions and Constraints:**
This User story is related to user story #20

rsyvulko:
Not relevant.

PR #116, opened 2019-09-03
#26 Access to cafe detail information

DmytroDovhal:


src/main/java/greencity/dto/place/PlaceInfoDto.java
```diff
@@ -13,10 +13,11 @@
...
 @AllArgsConstructor
 public class PlaceInfoDto {
+    private Long id;
     private String name;
     private LocationDto location;
     private List<OpenHoursDto> openingHoursList;
     private List<SpecificationValueDto> specificationValues;
     private List<CommentDto> comments;
-    private Byte rate;
+    private double rate;
```
tarasgl: Use Double object not primitive

src/main/java/greencity/repository/PlaceRepo.java
```diff
@@ -29,7 +29,7 @@
      * @return average rate
      */
     @Query(value = "select avg(r.rate) FROM Rate r " + "where place_id = :id")
-    Byte averageRate(@Param("id") Long id);
+    double averageRate(@Param("id") Long id);
```
tarasgl: Same here (Double object)

src/test/java/greencity/service/impl/PlaceServiceImplTest.java
```diff
@@ -114,44 +112,30 @@ private void setPlaceToLocalionTest(Place expectedPlace) {
...
-
-        assertNotNull(foundList);
-        for (AdminPlaceDto dto : foundList) {
-            assertEquals(dto.getStatus(), PlaceStatus.PROPOSED);
-            log.info(dto.toString());
-        }
+        List<Place> gen = new ArrayList<>();
+        List<AdminPlaceDto> genDto = new ArrayList<>();
+        List<AdminPlaceDto> res = new ArrayList<>();
+        assertEquals(res, genDto);
```
tarasgl: This test is useless, you are comparing two empty lists
Either delete it or if you have time make it as it should be

PR #123, opened 2019-09-03
#43 add list of cafes near the map

MarianMilian:


src/main/java/greencity/exception/CustomExceptionHandler.java
```diff
@@ -17,69 +17,104 @@
...
+     * @return ResponseEntity witch  contain http status and body  with message of exception.
+     * @author Nazar Stasyuk
      */
     @ExceptionHandler(BadEmailException.class)
-    public final ResponseEntity handle(BadEmailException e, WebRequest request) {
+    public final ResponseEntity handle(BadEmailException ex) {
         ValidationExceptionDto validationExceptionDto =
-            new ValidationExceptionDto("email", e.getMessage());
-        log.trace(e.getMessage(), e);
+            new ValidationExceptionDto("email", ex.getMessage());
```
Pitcity: "email" is a key for error? 
if so better to have separate constant or enum for keys.

Check out other places where ValidationExceptionDto is used.

> .body(Collections.singletonList(validationExceptionDto));

PR #127, opened 2019-09-03
Heroku deploy

RomanZahorui:


src/main/resources/liquibase.properties
```diff
@@ -1,5 +1,5 @@
-url=jdbc:mysql://localhost:3306/GreenCity?useUnicode=true&useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=UTC
-username=root
-password=root
+url=jdbc:mysql://szcsj0mhs2cjdwza:flnbp0iv4f13o8qy@l3855uft9zao23e2.cbetxkdyhwsb.us-east-1.rds.amazonaws.com:3306/arw5w15hky0z3d0t
```
tarasgl: You can read this property as an environment variable also using ${JAWSDB_URL}

Issue #178, opened 2019-09-18
BUG (#29) criteria for password does not work


ghost:
expected result:
when registering, user cannot register with a password that does not meet criteria as described in #29 acc 3

actual result:
user with email "masterlviv@gmail.com" successfully registered with password that does not meet the above stated criteria
![image](https://user-images.githubusercontent.com/53868652/65165633-dace5800-da47-11e9-9922-03389352e68c.png)

PR #181, opened 2019-09-19
#178 bug criteria for password does not work

GuN41K:


src/main/java/greencity/dto/filter/FilterDistanceDto.java
```diff
@@ -0,0 +1,22 @@
...
+import javax.validation.constraints.Positive;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class FilterDistanceDto {
+    @Min(value = -90, message = "Has to be greatest or equals -90")
```
MarianMilian: message  should be locate  in static constant

Issue #156, opened 2019-09-16
Filtering place by distance from user


GuN41K:
I have to calculate distance from user for each place, show it in sidebar and place info window, allow user to do filtering by it.
Parent: #153 
[Repo.](https://github.com/SoftServe-Social-Projects/GreenCityClient/tree/%23156-Filtering-places-by-distance-from-user)

PR #188, opened 2019-09-20
#156 filtering places by distance from user

GuN41K:


src/main/java/greencity/dto/filter/FilterDistanceDto.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class FilterDistanceDto {
+    @Min(value = -90, message = LAT_MIN_VALIDATION)
+    @Max(value = 90, message = LAT_MAX_VALIDATION)
+    private Double lat;
+    @Min(value = -180, message = LNG_MIN_VALIDATION)
```
Pitcity: is that distance will be fixed all the time or it might vary in the future? if so its better to have constants for those values so to use it both in validation and in messages combining

src/main/java/greencity/service/impl/PlaceServiceImpl.java
```diff
@@ -250,8 +251,41 @@ public Double averageRate(Long id) {
...
+            && distanceFromUserDto.getLat() != null
+            && distanceFromUserDto.getLng() != null
+            && distanceFromUserDto.getDistance() != null) {
+            placeList = placeList.stream().filter(place -> {
+                double userLatRad = Math.toRadians(distanceFromUserDto.getLat());
+                double userLngRad = Math.toRadians(distanceFromUserDto.getLng());
+                double placeLatRad = Math.toRadians(place.getLocation().getLat());
+                double placeLngRad = Math.toRadians(place.getLocation().getLng());
+
+                double distance = 6371 * Math.acos(
```
Pitcity: what is "6371"?

GuN41K: It is a constant from formula of Haversine. If you want to get distance in KM, you need to use 6371 constant. If you want to get in miles, replace it by 3959.

Issue #91, opened 2019-08-27
Editing existent place in the list as Admin 


nhrab:
As Admin I want to edit existent place so that information in the list of Places is always actual.

Mockup:
![image](https://user-images.githubusercontent.com/53810790/63801911-19c22f80-c91a-11e9-9845-cfb9ca5eb8ad.png)

Acceptance criteria:

1. "Edit" icon button is available on the "Places" tab near each row on the right side. It is shown as icon of pencil.

2. WHEN Admin see "Places" tab, THEN he can click on one of "Edit" icon button near each row.

3. WHEN Admin clicks "Edit" icon button THEN popup with already filled information of chosen row is shown:
   a. input field "Name"
   b. input field "Address"
   c. input field "Working times"
   d. "Save" button

5. WHEN Admin updates information in required fields of current place and clicks "Save" button THEN he can see this place in the list with updated info.

rsyvulko:
Not relevant.

PR #189, opened 2019-09-21

rinagoroh:


src/main/java/greencity/entity/OpeningHours.java
```diff
@@ -3,18 +3,16 @@
...
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-import lombok.NoArgsConstructor;
+import lombok.*;
 import org.springframework.format.annotation.DateTimeFormat;
 
 @Entity
-@Data
+@Getter
```
MarianMilian: Why  annotation  Data changed  to Getter and Setter anotations

src/main/java/greencity/service/OpenHoursService.java
```diff
@@ -56,4 +57,19 @@
...
+     * @return a list of the {@code OpeningHours} for the place by id.
+     */
+    Set<OpeningHours> findAllByPlaceId(Long placeId);
+
+    /**
+     * Delete all {@code OpeningHours} records related to the specified {@code Place}.
+     *
+     * @param placeId to find by.
+     */
+    void deleteAllByPlaceId(Long placeId);
```
ZakharSkaletskyi: Why it return nothing? We are making delete methods with anything in return statement, for example count of deleted rows.

src/main/java/greencity/entity/OpeningHours.java
rinagoroh: because I had exception:((

PR #207, opened 2019-09-24

rinagoroh:


src/main/java/greencity/service/impl/PlaceServiceImpl.java
```diff
@@ -78,87 +76,100 @@ public Place save(PlaceAddDto dto, String email) {
         Category category = categoryService.findByName(dto.getCategory().getName());
         Place place = modelMapper.map(dto, Place.class);
 
-        place.setAuthor(userService.findByEmail(email).orElseThrow(
-            () -> new NotFoundException(ErrorMessage.USER_NOT_FOUND_BY_EMAIL)));
+        getUserByEmailAndSetToPlace(email, place);
```
Pitcity: better to rename like "setUserToPlaceByEmail"

src/main/java/greencity/service/impl/PlaceServiceImpl.java
```diff
@@ -78,87 +76,100 @@ public Place save(PlaceAddDto dto, String email) {
...
+        place.setCategory(category);
+        place.setLocation(modelMapper.map(dto.getLocation(), Location.class));
```
Pitcity: so the location is saving now when you save place?
but why can't you use the same approach to Discounts and Opening hours so that you use separate "save"?

insert.sql
```diff
@@ -167,23 +166,15 @@ values ('2019-05-11', 'good', 4, 1);
...
-INSERT INTO location (address, lat, lng, place_id)
-VALUES ('Вулиця Кульпарківська, 226а, Львів, Львівська область, 79000', 49.807129, 23.977985, 2);
-INSERT INTO location (address, lat, lng, place_id)
-VALUES ('Площа Ринок, Львів, Львівська область, 79000', 49.842042, 24.030359, 3);
-INSERT INTO location (address, lat, lng, place_id)
-VALUES ('Проспект В''ячеслава Чорновола, 2, Львів, Львівська область, 79000', 49.847489, 24.025975, 4);
-INSERT INTO location (address, lat, lng, place_id)
-VALUES ('Площа Ринок, 14 (підвал), Львів, Львівська область, 79000', 49.841311, 24.03229, 5);
-insert into discount
-values (1, 3, 1, 1, 2),
```
MarianMilian: why you  delete discount inserts?

rinagoroh: my mistake

PR #244, opened 2019-10-03

rinagoroh:


src/main/java/greencity/dto/specification/SpecificationNameDto.java
```diff
@@ -5,11 +5,13 @@
...
+import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 
 @Data
 @NoArgsConstructor
 @AllArgsConstructor
+@EqualsAndHashCode(exclude = {"name"})
 public class SpecificationNameDto {
     @NotNull(message = ValidationConstants.EMPTY_SPECIFICATION_NAME)
     @NotEmpty(message = ValidationConstants.EMPTY_SPECIFICATION_NAME)
```
GuN41K: @NotNull is inside @NotEmpty annotation, so you need not use @NotNull here. But also you should thing about @NotBlank annotation.

src/main/java/greencity/dto/discount/DiscountValueDto.java
```diff
@@ -4,19 +4,17 @@
...
 @AllArgsConstructor
-@Builder
-public class DiscountDto {
-    @Range(min = ValidationConstants.DISCOUNT_VALUE_MIN,
-        max = ValidationConstants.DISCOUNT_VALUE_MAX)
-    private int value;
+public class DiscountValueDto {
+    @Range(min = 0, message = ValidationConstants.DISCOUNT_VALUE_MIN_ERROR)
+    @Range(max = 100, message = ValidationConstants.DISCOUNT_VALUE_MAX_ERROR)
+    private int value = 0;
```
GuN41K: @Range has "max" and "min" value, so you can past here single annotation with one error message like "has to be more than 0 and less than 100"

rinagoroh: ok

src/main/java/greencity/repository/options/PlaceFilter.java
```diff
@@ -126,11 +126,9 @@ private Predicate hasDiscount(Root<Place> r, CriteriaBuilder cb, FilterDiscountD
             return cb.conjunction();
         }
         return cb.and(
-            cb.equal(r.join("discounts").join("category").get("name"),
-                discount.getCategory().getName()),
-            cb.equal(r.join("discounts").join("specification").get("name"),
+            cb.equal(r.join("discountValues").join("specification").get("name"),
```
ZakharSkaletskyi: instead of strings there should be constants

src/main/java/greencity/service/impl/PlaceServiceImpl.java
```diff
@@ -149,41 +118,37 @@ public Place update(PlaceUpdateDto dto) {
...
 
     /**
-     * Method for updating set of {@link Discount} and save with new {@link Category} and {@link Place}.
+     * Method for updating set of {@link DiscountValue} and save with new {@link Category} and {@link Place}.
      *
-     * @param discountDtos    - set of {@link Discount}.
-     * @param updatedCategory - {@link Category} entity.
-     * @param updatedPlace    - {@link Place} entity.
+     * @param discounts    - set of {@link DiscountValue}.
+     * @param updatedPlace - {@link Place} entity.
```
ZakharSkaletskyi: You should write javadoc in one way - always {@inheritDoc} or always write all doc info

src/main/java/greencity/service/impl/PlaceServiceImpl.java
```diff
@@ -149,41 +118,37 @@ public Place update(PlaceUpdateDto dto) {
...
+            DiscountValue discount = discountValueMapper.convertToEntity(d);
             discount.setPlace(updatedPlace);
-            discount.setCategory(updatedCategory);
-            discount.setSpecification(specification);
             discountService.save(discount);
-            discounts.add(discount);
+            newDiscounts.add(discount);
         });
-        discountsOld.addAll(discounts);
+        discountsOld.addAll(newDiscounts);
```
ZakharSkaletskyi: Why you add to discountsOld newDiscounts, but nothing do later with discountsOld? And what you do with newDiscounts? You only add it to variable discountsOld and then method ends.

src/main/java/greencity/service/impl/PlaceServiceImpl.java
```diff
@@ -380,6 +345,7 @@ public Double averageRate(Long id) {
     @Override
     public List<PlaceByBoundsDto> getPlacesByFilter(FilterPlaceDto filterDto) {
         List<Place> list = placeRepo.findAll(new PlaceFilter(filterDto));
+        log.error(list.toString());
```
ZakharSkaletskyi: why you add to error level all elements from list?

src/main/java/greencity/service/impl/SpecificationServiceImpl.java
```diff
@@ -83,6 +83,10 @@ public Long deleteById(Long id) {
      */
     @Override
     public Specification findByName(String name) {
+        Specification specification = specificationRepo.findByName(name);
+        if (specification == null) {
+            throw new NotFoundException(ErrorMessage.SPECIFICATION_NOT_FOUND_BY_NAME + name);
+        }
         return specificationRepo.findByName(name);
```
ZakharSkaletskyi: you make 2 DB calls. You should return "specification" variable, but not do second request to DB.

src/main/resources/application-dev.properties
```diff
@@ -12,7 +12,7 @@ spring.mail.password=${password}
 spring.mail.properties.mail.smtp.auth=true
 spring.mail.properties.mail.smtp.starttls.enable=true
 verifyEmailTimeHour=24
-spring.liquibase.enabled=true
+spring.liquibase.enabled=false
```
ZakharSkaletskyi: Why you disable liquibase?

src/test/java/greencity/service/impl/PlaceServiceImplTest.java
```diff
@@ -91,11 +90,11 @@
...
-    Discount discountEntity = Discount.builder()
-        .id(1L)
-        .category(category)
-        .specification(specificationEntity)
-        .build();
+//    Discount discountEntity = Discount.builder()
+//        .id(1L)
+//        .category(category)
+//        .specification(specificationEntity)
+//        .build();
```
ZakharSkaletskyi: delete it

src/test/java/greencity/service/impl/PlaceServiceImplTest.java
```diff
@@ -76,9 +75,9 @@
 
     Set<OpeningHoursDto> openingHoursList = new HashSet<>();
 
-    Set<Discount> discountEntities = new HashSet<>();
+//    Set<Discount> discountEntities = new HashSet<>();
 
-    Set<DiscountDto> discountDtos = new HashSet<>();
+    // Set<DiscountDto> discountDtos = new HashSet<>();
```
ZakharSkaletskyi: delete it

src/test/java/greencity/service/impl/PlaceServiceImplTest.java
```diff
@@ -139,8 +138,8 @@
     @Mock
     private SpecificationService specificationService;
 
-    @Mock
-    private DiscountService discountService;
+//    @Mock
+//    private DiscountService discountService;
```
ZakharSkaletskyi: delete

src/main/java/greencity/service/impl/PlaceServiceImpl.java
rinagoroh: be more attentive! it's a private method

PR #247, opened 2019-10-03

rinagoroh:


src/main/java/greencity/service/impl/PlaceServiceImpl.java
```diff
@@ -149,41 +118,37 @@ public Place update(PlaceUpdateDto dto) {
...
+            DiscountValue discount = discountValueMapper.convertToEntity(d);
             discount.setPlace(updatedPlace);
-            discount.setCategory(updatedCategory);
-            discount.setSpecification(specification);
             discountService.save(discount);
-            discounts.add(discount);
+            newDiscounts.add(discount);
         });
-        discountsOld.addAll(discounts);
+        discountsOld.addAll(newDiscounts);
```
ZakharSkaletskyi: Why you add to discountsOld newDiscounts, but nothing do later with discountsOld? And what do you do with newDiscounts? You only add it to method's variable discountsOld and then method ends.

src/test/java/greencity/service/impl/PlaceServiceImplTest.java
```diff
@@ -91,20 +86,14 @@
...
-
     Place place = Place.builder()
         .id(1L)
         .name("Test")
         .category(category)
         .author(user)
         .location(location)
         .openingHoursList(openingHoursListEntity)
-        .discounts(discountEntities)
+//        .discounts(discountEntities)
```
ZakharSkaletskyi: delete it

src/main/java/greencity/repository/options/PlaceFilter.java
```diff
@@ -126,11 +126,9 @@ private Predicate hasDiscount(Root<Place> r, CriteriaBuilder cb, FilterDiscountD
             return cb.conjunction();
         }
         return cb.and(
-            cb.equal(r.join("discounts").join("category").get("name"),
-                discount.getCategory().getName()),
-            cb.equal(r.join("discounts").join("specification").get("name"),
+            cb.equal(r.join("discountValues").join("specification").get("name"),
                 discount.getSpecification().getName()),
-            cb.between(r.join("discounts").get("value"),
+            cb.between(r.join("discountValues").get("value"),
```
ZakharSkaletskyi: Instead of strings there should be constants. In PlaceFilter many strings with the same name.

PR #258, opened 2019-10-10
fix code smells and save method in FavoritePlace

ZakharSkaletskyi:


src/main/java/greencity/constant/ErrorMessage.java
```diff
@@ -5,8 +5,9 @@
     public static final String USER_NOT_FOUND_BY_EMAIL = "The user does not exist by this email: ";
     public static final String USER_WITH_EMAIL_EXIST = "The user is already exist by this email: ";
     public static final String PLACE_NOT_FOUND_BY_ID = "The place does not exist by this id: ";
-    public static final String FAVORITE_PLACE_NOT_FOUND = "The favorite place does not exist by this placeId: ";
-    public static final String FAVORITE_PLACE_ALREADY_EXISTS = "Favorite place already exist for this place";
+    public static final String FAVORITE_PLACE_NOT_FOUND = "The favorite place does not exist ";
+    public static final String FAVORITE_PLACE_ALREADY_EXISTS = "Favorite place already exist for this placeId:  %d and "
+        + "user with email: %s";
```
nazarvladyka: You can leave it in 1 line, checkstyle is OK with it

PR #264, opened 2019-10-15
#158 add feedback to a place

MarianMilian:


src/main/java/greencity/controller/PlaceCommentController.java
```diff
@@ -0,0 +1,74 @@
...
+        @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
+    })
+    @PostMapping("/place/{placeId}/comments")
+    public ResponseEntity save(@PathVariable Long placeId,
+                               @Valid @RequestBody AddCommentDto addCommentDto, Principal principal) {
+        User user = userService.findByEmail(principal.getName())
+            .orElseThrow(() -> new NotFoundException(ErrorMessage.USER_NOT_FOUND_BY_EMAIL + principal.getName()));
+        if (user.getUserStatus().equals(UserStatus.BLOCKED)) {
+            throw new UserBlockedException(ErrorMessage.USER_HAS_BLOCKED_STATUS);
+        }
```
nazarvladyka: I am not sure, but I think all logic we should place at the service

nazarvladyka: Also you can do something like this 
image
instead of getting "principal" as a param

src/main/resources/application-dev.properties
```diff
@@ -22,7 +22,7 @@ address=http://localhost:8080
 client.address=http://localhost:4200
 google.clientId=${googleClientId}
 
-accessTokenValidTimeInMinutes=15
+accessTokenValidTimeInMinutes=1000
```
nazarvladyka: Return it back to 15

MarianMilian: done

src/main/java/greencity/controller/PlaceCommentController.java
MarianMilian: Principal is removed, this logic should located here as blocked user will not get access to service layer

PR #268, opened 2019-10-15

rinagoroh:


src/main/java/greencity/constant/ValidationConstants.java
```diff
@@ -65,7 +65,7 @@
     public static final String LNG_MAX_VALIDATION = "Has to be lower or equals 180";
 
     public static final String RATE_MIN_VALUE = "The rate must be at least {value}";
-    public static final String RATE_MAX_VALUE = "The rate must be less than {value}";
+    public static final String RATE_MAX_VALUE = "The rate must bigger than {min} and lover then {max}.";
```
RomanZahorui: lover means коханець

src/main/java/greencity/entity/Place.java
```diff
@@ -37,7 +37,7 @@
     @Column(unique = true, length = 50)
     private String email;
 
-    @OneToMany(mappedBy = "place", fetch = FetchType.LAZY, cascade = {CascadeType.PERSIST, CascadeType.MERGE})
+    @OneToMany(mappedBy = "place", cascade = {CascadeType.PERSIST, CascadeType.MERGE})
```
ZakharSkaletskyi: Why you delete FetchType.LAZY? You will get all comments every time, when you will get place from database.

src/main/java/greencity/constant/ValidationConstants.java
rinagoroh: ahahah it is not mine:) this code from Marian's branch

PR #313, opened 2019-11-04
added tests for OpenHoursService

yuraolkhovskyi:
added tests for OpenHoursService

### DIFF
--- a/src/test/java/greencity/service/impl/OpenHoursServiceImplTest.java
+++ b/src/test/java/greencity/service/impl/OpenHoursServiceImplTest.java
@@ -1,13 +1,11 @@
 package greencity.service.impl;
 
-import static com.sun.javaws.JnlpxArgs.verify;
 import static org.junit.Assert.assertEquals;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.when;
 
-import greencity.GreenCityApplication;
 import greencity.entity.BreakTime;
 import greencity.entity.OpeningHours;
 import greencity.entity.Place;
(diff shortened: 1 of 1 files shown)


src/test/java/greencity/service/impl/OpenHoursServiceImplTest.java
```diff
@@ -1,27 +1,31 @@
 package greencity.service.impl;
 
+import static com.sun.javaws.JnlpxArgs.verify;
```
ZakharSkaletskyi: Where you use this library?

src/test/java/greencity/service/impl/OpenHoursServiceImplTest.java
```diff
@@ -149,4 +145,35 @@ public void findAllTest() {
...
+        List<OpeningHours> genericOpeningHours =
+            new ArrayList<>(Arrays.asList(new OpeningHours(), new OpeningHours()));
+
+        when(openHoursRepo.findAllByPlace(any())).thenReturn(genericOpeningHours);
+
+        List<OpeningHours> foundOpeningHours = openHoursService.getOpenHoursByPlace(place);
+
+        assertEquals(genericOpeningHours, foundOpeningHours);
+        Mockito.verify(openHoursRepo, times(1)).findAllByPlace(any());
+    }
```
nazarvladyka: Please consider about removing some blank lines here and in **findAllByPlaceIdTest**(), now it's to much and it`s not good for big files where you will need to scroll all of this lines.
Consider about better grouping lines, for e.x:

Place place = Place.builder().build();
List<OpeningHours> genericOpeningHours = new ArrayList<>(Arrays.asList(new OpeningHours(), new OpeningHours()));
List<OpeningHours> foundOpeningHours = openHoursService.getOpenHoursByPlace(place);

 when(openHoursRepo.findAllByPlace(any())).thenReturn(genericOpeningHours);

 assertEquals(genericOpeningHours, foundOpeningHours);
 Mockito.verify(openHoursRepo, times(1)).findAllByPlace(any());

And at line **154** better to use just Arrays.asList(new OpeningHours(), new OpeningHours()), without new ArrayList()

src/test/java/greencity/service/impl/OpenHoursServiceImplTest.java
```diff
@@ -149,4 +145,35 @@ public void findAllTest() {
 
         assertEquals(genericEntities, foundEntities);
     }
+
+    @Test
+    public void getOpenHoursByPlaceTest() {
+        Place place = Place.builder().build();
```
GuN41K: What about keyword "new"?

yuraolkhovskyi: I think that there is no difference between creating an object using builder and constructor in such case but I refixed this to "new".

src/test/java/greencity/service/impl/OpenHoursServiceImplTest.java
yuraolkhovskyi: Actually I've done this in such way for more readability but I got you. And also when I put foundOpeningHours under genericOpeningHours test fails and the same case with set in method findAllByPlaceId.
Btw, thanks for useful hints.

PR #316, opened 2019-11-04
This is not a bug, but rather a Spring feature. When entered invalid …

yuriikoval1997:
This is not a bug, but rather a Spring feature. When entered invalid data, for instance, page index (less than 0) or size (less than 1), Spring uses its default Pageable object with page
equals to 0 and size equals to 20. I can override the default Pageable
object with different values, but I cannot make it throw an exception. We'd better leave as it is now and avoid creating a bicycle.

I modified ApiPageable annotation, added ApiResponce with code = 400 to
the corresponding controller and deleted unused import, created Pageable
configuration with max Pageable size = 100.

### DIFF
--- a/src/main/java/greencity/annotations/ApiPageable.java
+++ b/src/main/java/greencity/annotations/ApiPageable.java
@@ -11,9 +11,11 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 @ApiImplicitParams({
     @ApiImplicitParam(name = "page", dataType = "int", paramType = "query", defaultValue = "0",
-        value = "Results page you want to retrieve (0..N)"),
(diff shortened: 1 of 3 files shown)


GuN41K:
You can create your custom Pageable DTO with all validation that you need. After that you can map this DTO to "Spring Pageable" in your service and call repo with it.

yuriikoval1997:
> You can create your custom Pageable DTO with all validation that you need. After that you can map this DTO to "Spring Pageable" in your service and call repo with it. Read more about PageRequest object and his "of" method. P.S. PageRequest is subclass of Pageable

I have already tried that. It didn't work out. Spring replaces the invalid data before creating a Pageable object. You may have a look at resolveArgument method of class PageableHandlerMethodArgumentResolver.

PR #322, opened 2019-11-07
Fix bug update criteria method in user filter class

VolodjaTur:
Check please.


src/main/java/greencity/repository/options/UserFilter.java
```diff
@@ -52,11 +53,29 @@ public Predicate toPredicate(Root<User> root, CriteriaQuery<?> criteriaQuery, Cr
...
     }
+
+    /**
+     * Returns a String criteria for search.
+     *
+     * @param criteria  must not be {@literal null}.
+     * @return String creteria  may be {@literal null}.
+     */
+    private String replaceCriteria(String criteria) {
+        criteria = Optional.of(criteria).orElse("");
```
ZakharSkaletskyi: For what purpose do you use this assignment?
If criteria will have null value, you get NullPointerException. So you should check null value.

VolodjaTur: Check again please.

ZakharSkaletskyi: * @param criteria  must not be {@literal null}.
      * @return String creteria  may be {@literal null}
If criteria must not be null you shoud throw exception. If criteria can be null then you should change description.
And how can you return null if you assign criteria to a new value, when you get null as an argument?

ZakharSkaletskyi: If you want to use null check you can use criteria=Optional.ofNullable(criteria).orElse("");
Optional.of throw exception when argument null

PR #327, opened 2019-11-11
renamed names of tables in database according to convention where it'…

yuraolkhovskyi:
…s needed, update insert script, refactor custom queries, refactor entities


src/main/java/greencity/entity/DiscountValue.java
```diff
@@ -9,6 +9,7 @@
 @AllArgsConstructor
 @Builder
 @EqualsAndHashCode(exclude = {"value"})
+@Table(name = "DiscountValues")
```
nazarvladyka: should be "discount_values" according to other changes

PR #531, opened 2020-03-26
Firebase

juliayuvchenko:


core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,84 @@
...
+ * <p>This class is based from <a
+ * href="https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java">
+ * StorageExample.java</a>.
+ *
+ * <p>See the <a
+ * href="https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-examples/README.md">
+ * README</a> for compilation instructions. Run this code with
+ **/
+@Service
+public class CloudStorageServiceImpl implements FileService {
```
taraskovaliv: CloudStorageServiceImpl need to rename FileServiceImpl

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,84 @@
...
+            final String blob = UUID.randomUUID().toString();
+            final BlobInfo blobInfo = BlobInfo.newBuilder(bucketName, blob).setContentType(contentType).build();
+
+            return run(storage, blobInfo, multipartFile);
+        } catch (IOException ex) {
+            throw ex;
+        }
+    }
+
+    private static URL run(final Storage storage,
```
taraskovaliv: maybe method name save or write will be suitable

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Upload file to Google Cloud Storage.
+ *
+ * <p>This class is based from <a
+ * href="https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/StorageExample.java">
+ * StorageExample.java</a>.
+ *
+ * <p>See the <a
+ * href="https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-examples/README.md">
+ * README</a> for compilation instructions. Run this code with
```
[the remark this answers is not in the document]
juliayuvchenko: Checkstyle doesn't pass without this javadoc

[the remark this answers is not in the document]
juliayuvchenko: okay, I will remove it.

core/src/main/java/greencity/controller/FileServiceController.java
```diff
@@ -45,6 +47,7 @@ public FileServiceController(FileService fileService) {
         @ApiResponse(code = 201, message = HttpStatuses.CREATED, response = EcoNews.class),
         @ApiResponse(code = 303, message = HttpStatuses.SEE_OTHER),
         @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
+        @ApiResponse(code = 500, message = HttpStatuses.INTERNAL_SERVER_ERROR),
     })
     @PostMapping("/image")
     public ResponseEntity<String> uploadImage(@RequestParam("Image to save") @NotEmpty MultipartFile multipartFile,
```
vterlyha: FIle for sure should be send in body, so it should be @RequestBody or @RequestPart annotation. It looks like you haven't test it - because you won't be able to send file in URL

vterlyha: Also bad naming - use cammelCase

core/src/main/java/greencity/controller/FileServiceController.java
```diff
@@ -53,6 +56,10 @@ public FileServiceController(FileService fileService) {
                                               @RequestParam("folder name")
```
vterlyha: same her about cammelCase

core/src/main/java/greencity/service/FileService.java
```diff
@@ -1,17 +1,18 @@
...
-     *
-     * @param multipartFile   file for saving.
-     * @param imageFolderName name of folder in which the photo will be saved.
-     * @return url of the saved photo.
-     */
-    String uploadImage(MultipartFile multipartFile, String imageFolderName);
+     * Upload file to Google Cloud Storage.
+     * @param multipartFile image file to save.
+     * @return public image url.
+     * @throws IllegalArgumentException exception.
```
vterlyha: I don't see this method throws IllegalArgumentException

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+                                   @Value("${BUCKET_NAME}") final String bucketName) {
+        this.projectId = projectId;
+        this.bucketName = bucketName;
+        this.optionsBuilder = StorageOptions.newBuilder();
+        if (bucketName == null || bucketName.isEmpty()) {
+            throw new IllegalArgumentException("Bucket name is null");
+        }
+        if (projectId != null && !projectId.isEmpty()) {
+            optionsBuilder.setProjectId(projectId);
+        }
```
vterlyha: BucketName and ProjectID can't be null because you are taking them from property file

vterlyha: Also, please use Lombok annotation @RequiredArgsConstructor for creating constructors. Write constructor by yourself only if you need specific one

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+    public URL upload(final MultipartFile multipartFile) throws IOException {
+        try {
+            final String contentType = multipartFile.getContentType();
+            final String blob = UUID.randomUUID().toString();
+            final BlobInfo blobInfo = BlobInfo.newBuilder(bucketName, blob).setContentType(contentType).build();
+
+            return write(storage, blobInfo, multipartFile);
+        } catch (IOException ex) {
+            throw ex;
+        }
```
vterlyha: Do you really need to handle IOException ?

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+        this.storage = optionsBuilder.build().getService();
+    }
+
+
+    @Override
+    public URL upload(final MultipartFile multipartFile) throws IOException {
+        try {
+            final String contentType = multipartFile.getContentType();
+            final String blob = UUID.randomUUID().toString();
```
vterlyha: This variable should be named 'key', as I understand.

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
vterlyha: And you don't need to use final for every variable. Please, use it only when you explicitly don't want it to be changed

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+            final BlobInfo blobInfo = BlobInfo.newBuilder(bucketName, blob).setContentType(contentType).build();
+
+            return write(storage, blobInfo, multipartFile);
```
vterlyha: No need to pass storage here - it is you instance variable

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+        try (WriteChannel writer = storage.writer(blobInfo)) {
+            final byte[] buffer = new byte[1024];
+            try (InputStream input = multipartFile.getInputStream()) {
+                int limit;
+                while ((limit = input.read(buffer)) >= 0) {
+                    try {
+                        writer.write(ByteBuffer.wrap(buffer, 0, limit));
+                    } catch (Exception ex) {
+                        ex.printStackTrace();
+                    }
```
vterlyha: Do you need to catch exception here. Also, catching Exception and not a specific child is a bad practice

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private static URL write(final Storage storage,
+                             final BlobInfo blobInfo, final MultipartFile multipartFile) throws IOException {
+        try (WriteChannel writer = storage.writer(blobInfo)) {
+            final byte[] buffer = new byte[1024];
+            try (InputStream input = multipartFile.getInputStream()) {
+                int limit;
+                while ((limit = input.read(buffer)) >= 0) {
+                    try {
+                        writer.write(ByteBuffer.wrap(buffer, 0, limit));
```
vterlyha: This part should be rewritten. I'm sure there is some libraries you can use to save file to google cloud without 3 try block, where all this logic is encapsulated.

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+                    } catch (Exception ex) {
+                        ex.printStackTrace();
+                    }
+                }
+            }
+        }
+
+        return new URL("https://storage.cloud.google.com/" + blobInfo.getBucket()
+            + "/" + blobInfo.getName() + "?authuser=0");
```
vterlyha: String concatenation is a bad practice

vterlyha: "https://storage.cloud.google.com/" - this one should be taken to properties file or at least constant.

vterlyha: ?authuser=0 - what is this ?

core/src/main/java/greencity/service/impl/CloudStorageServiceImpl.java
juliayuvchenko: The reason I use final everywhere in my code:
1. This improves readability of the code;
2. Code is better structured;
3. It gives hints JVM to produced optimized CPU instructions.

juliayuvchenko: https://stackoverflow.com/questions/316352/why-would-one-mark-local-variables-and-method-parameters-as-final-in-java

core/src/main/java/greencity/controller/FileServiceController.java
```diff
@@ -36,23 +32,18 @@ public FileServiceController(FileService fileService) {
...
     })
     @PostMapping("/image")
-    public ResponseEntity<String> uploadImage(@RequestParam("Image to save") @NotEmpty MultipartFile multipartFile,
-                                              @ApiParam(value = "Folder name where the image will be saved('eco_news') "
-                                                  + "The maximum size of an image is not more than 2 mb.")
-                                              @RequestParam("folder name")
-                                              @Pattern(regexp = VALIDATION_FOLDER, message = "invalid folder name")
-                                              @NotNull(message = "can not be null") String folderName) {
-        return ResponseEntity.status(HttpStatus.OK).body(fileService.uploadImage(multipartFile, folderName));
+    public ResponseEntity<String> uploadImage(@RequestParam @NotEmpty MultipartFile image) {
```
vterlyha: @RequestParam ? It should be @RequestBody

PR #532, opened 2020-03-30
Add image to add eco news request

taraskovaliv:


core/src/test/java/greencity/TestConst.java
```diff
@@ -0,0 +1,8 @@
+package greencity;
+
+public final class TestConst {
+    public static final String site = "https://google.com/";
+    public static final String name = "Taras";
+    public static final String email = "taras@gmail.com";
+    public static final String imgName = "Screenshot.png";
```
ViktoriyaRyazhska: According to code convention constants should be named in upper case, e.g SITE or IMAGE_NAME

PR #615, opened 2020-05-04
[Minor Bug-Fix] Temporary bug fix when user sign in

zendermean:
Deleting the logic of adding standard habits to the user at first login.


core/src/main/java/greencity/security/service/impl/OwnSecurityServiceImpl.java
```diff
@@ -32,6 +29,9 @@
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
+import static greencity.constant.ErrorMessage.*;
+import static greencity.constant.RabbitConstants.VERIFY_EMAIL_ROUTING_KEY;
```
vterlyha, line 33: Do you need these imports ?

PR #630, opened 2020-05-06
Junit tests for SearchServiceImpl, HabitServiceImpl, NotificationServiceImpl

MarDatsko:
added Junit tests for SearchServiceImpl, HabitServiceImpl, NotificationServiceImpl

### DIFF
--- a/core/src/test/java/greencity/service/impl/HabitServiceImplTest.java
+++ b/core/src/test/java/greencity/service/impl/HabitServiceImplTest.java
@@ -30,11 +30,10 @@ class HabitServiceImplTest {
     @Mock
     private HabitRepo habitRepo;
 
-    private Habit habit = ModelUtils.getHabit();
-    private HabitDictionaryTranslation habitDictionaryTranslation = ModelUtils.getHabitDictionaryTranslation();
-
     @Test
     void getHabitDictionaryTranslation() {
+        Habit habit = ModelUtils.getHabit();
+        HabitDictionaryTranslation habitDictionaryTranslation = ModelUtils.getHabitDictionaryTranslation();
         when(habitDictionaryTranslationRepo.findByHabitDictionaryAndLanguageCode(any(HabitDictionary.class), anyString()))
                 .thenReturn(Optional.of(habitDictionaryTranslation));
         assertEquals(habitDictionaryTranslation, habitService.getHabitDictionaryTranslation(habit, "en"));
@@ -42,6 +41,7 @@ class HabitServiceImplTest {
 
     @Test
     void getById() {
+        Habit habit = ModelUtils.getHabit();
         when(habitRepo.findById(anyLong())).thenReturn(Optional.of(habit));
         assertEquals(habit, habitService.getById(1L));
     }
--- a/core/src/test/java/greencity/service/impl/HabitServiceImplTest.java
+++ b/core/src/test/java/greencity/service/impl/HabitServiceImplTest.java
@@ -1,0 +1,48 @@
+package greencity.service.impl;
+
+import greencity.ModelUtils;
+import greencity.entity.Habit;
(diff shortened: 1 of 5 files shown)


core/src/test/java/greencity/service/impl/HabitServiceImplTest.java
```diff
@@ -0,0 +1,48 @@
...
+    private HabitServiceImpl habitService;
+
+    @Mock
+    private HabitDictionaryTranslationRepo habitDictionaryTranslationRepo;
+
+    @Mock
+    private HabitRepo habitRepo;
+
+    private Habit habit = ModelUtils.getHabit();
+    private HabitDictionaryTranslation habitDictionaryTranslation = ModelUtils.getHabitDictionaryTranslation();
```
vterlyha, line 34: Good approach to create testing elements inside each method in case you will need to make different changes when new tests will be added

core/src/test/java/greencity/service/impl/NotificationServiceImplTest.java
```diff
@@ -1,114 +1,154 @@
...
-        testPlaces = Arrays.asList(testPlace1, testPlace2);
-
-        when(placeRepo.findAllByModifiedDateBetweenAndStatus(
-            any(LocalDateTime.class),
-            any(LocalDateTime.class),
-            eq(PlaceStatus.APPROVED)
-        )).thenReturn(testPlaces);
+    private EmailNotification emailNotification;
+    private Category category = ModelUtils.getCategory();
+    private User user = ModelUtils.getUser();
```
vterlyha, line 52: Same here

core/src/test/java/greencity/service/impl/NotificationServiceImplTest.java
```diff
@@ -1,114 +1,154 @@
...
+
+    @BeforeEach
+    void setUp() {
```
vterlyha, line 55: This method is needed to make some configuration, for example like autowiring of some classes or making additional configurations and not for mocking data. Please, move it to appropriate methods

core/src/test/java/greencity/service/impl/NotificationServiceImplTest.java
```diff
@@ -1,114 +1,154 @@
...
-        verify(userRepo).findAllByEmailNotification(emailNotification);
+        verify(rabbitTemplate, Mockito.times(1))
+                .convertAndSend(any(), anyString(), any(SendReportEmailMessage.class));
+
     }
 
     @Test
-    public void sendMonthlyReportTest() {
+    void sendMonthlyReportTest() {
         emailNotification = EmailNotification.MONTHLY;
```
vterlyha, line 140: Same here - create testing object inside the method

core/src/test/java/greencity/service/impl/SearchServiceImplTest.java
```diff
@@ -0,0 +1,40 @@
...
+    private SearchServiceImpl searchService;
+
+    @InjectMocks
+    private ModelMapper modelMapper;
+
+    @Mock
+    private EcoNewsService ecoNewsService;
+
+    @Test
+    void searchTest() {
```
vterlyha, line 33: Bad test. Nothing is tested here. You new to mock repo not service

MarDatsko: Maybe I something not understands. 
`SearchServiceImpl.class` have only one field with `EcoNewsService`  and one method `search`, which only created a new object `SearchResponseDto`, and I in test mock this services ( `EcoNewsService` ) and check if the object is created correctly.

I added a new test for `EcoNewsServiceImpl` where I mock `EcoNewsRepo`, and checked if we correctly map object what get from DB;

PR #732, opened 2020-05-19
Added daily, weekly and monthly email notifications about not marking…

WayUp06:


core/src/main/java/greencity/config/ScheduleConfig.java
```diff
@@ -11,5 +28,61 @@
...
+    private final RabbitTemplate rabbitTemplate;
+    private final UserRepo userRepo;
+
+    /**
+     * Invoke {@link sendHabitNotification} from EmailMessageReceiver to send email letters
+     * to each user that hasn't marked any habit during last 3 days.
+     *
+     * @param users     list of potential {@link User} to send notifications.
+     */
+    public void sendHabitNotificationIfNeedBe(List<User> users) {
```
vterlyha, line 43: Is there a mistake in the end of the naming ?

vterlyha: Maybe this method can be private ?

PR #752, opened 2020-05-25
Profile picture for User

MarDatsko:
Added ability to save profile picture for user


core/src/main/java/greencity/service/impl/UserServiceImpl.java
```diff
@@ -92,6 +88,11 @@
     private final GoalTranslationRepo goalTranslationRepo;
     private final HabitDictionaryService habitDictionaryService;
     private final HabitDictionaryTranslationRepo habitDictionaryTranslationRepo;
+    private final FileService fileService;
+
+    private final String defaultProfilePicture = "https://storage.cloud.google.com"
+            + "/staging.greencity-c5a3a.appspot.com"
+            + "/d333665a-9269-49ef-bc77-2d8d4090290f";
```
vterlyha, line 95: Move this to properties file

core/src/main/java/greencity/service/impl/UserServiceImpl.java
```diff
@@ -630,4 +634,18 @@ public void addDefaultHabit(Long userId, String language) {
...
         return habitDictionaryDtos;
     }
+
+    /**
+     * Get profile picture path {@link String}.
+     *
+     * @return profile picture path {@link String}
+     */
+    @Override
+    public String getProfilePicturePathByUserId(Long id) {
```
vterlyha, line 644: Where is this method used ? Maybe you shouldn't throw exception if picture path not found as user signs in for the first time - he won't have profile picture ?

PR #757, opened 2020-05-27
Profile picture with the changed files

MarDatsko:


core/src/main/java/greencity/constant/AppConstant.java
```diff
@@ -13,6 +13,7 @@
     public static final String VALIDATION_EMAIL = "^[a-zA-Z0-9_!#$%&'*+/=?`{|}~^.-]+@[a-zA-Z0-9.-]+$";
     public static final String DEFAULT_LANGUAGE_CODE = "en";
     public static final String VALIDATION_FOLDER = "(^eco_news$)";
+    public static final String DEFAULT_PROFILE_PICTURE = "https://storage.cloud.google.com/staging.greencity-c5a3a.appspot.com\"/d333665a-9269-49ef-bc77-2d8d4090290f";
```
zendermean, line 16: I think, better should be move this to environment variables.

PR #872, opened 2020-06-10
User friends

MarDatsko:
User friends


core/src/main/java/greencity/controller/UserController.java
```diff
@@ -575,19 +578,87 @@ public void deleteHabit(
...
+
+    /**
+     * Method for deleting user friend.
+     *
+     * @param friendId id user friend.
+     * @param userId   id current user.
+     * @author Marian Datsko
+     */
+    @ApiOperation(value = "Delete user friend")
+    @ApiResponses(value = {
```
antonbakalets, line 605: [optional] Why are you adding those `ApiResponses `with standard messages? 
It makes sense to add them if you want inform user of something with the `message` param.
I know that 200 - it's OK, that 400 - it's bad request. Those additional lines of code doesn't give any additional value.

core/src/main/java/greencity/repository/UserRepo.java
```diff
@@ -79,4 +79,37 @@
      */
     @Query("SELECT profilePicturePath FROM User WHERE id=:id")
     Optional<String> getProfilePicturePathByUserId(Long id);
+
+    /**
+     * Get all user friends{@link User}.
+     *
+     * @return list of {@link User}.
+     */
+    @Query(value = " SELECT * FROM public.users_friends "
```
antonbakalets, line 88: Why this query is written in native SQL? I would prefer JPQL. Reasoning: JPQL is independent of database server dialect. Event SQL is a standard each server (Postgres, Oracle, mySql) have dialects. The same bellow.

antonbakalets, line 88: Please remove schema name `public.` from the query. The same bellow.

core/src/main/java/greencity/repository/UserRepo.java
```diff
@@ -79,4 +79,37 @@
...
+        + " LEFT JOIN public.users ON public.users.id = public.users_friends.friends "
+        + " WHERE public.users_friends.user_id = :userId ", nativeQuery = true)
+    List<User> getAllUserFriends(Long userId);
+
+    /**
+     * Delete friend {@link User}.
+     */
+    @Modifying
+    @Query(value = "DELETE FROM public.users_friends WHERE user_id= :userId AND friends= :friendId", nativeQuery = true)
```
antonbakalets, line 97: I usually put `nativeQuery = true` at the beginning, so when I'm reading the code I understand from the start which notation is used.

core/src/main/resources/db/changelog/logs/ch-users-friends-Datsko.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet author="Marian Datsko" id="1571227585316-98">
+        <createTable tableName="users_friends">
+            <column name="user_id" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="friends" type="BIGINT">
```
antonbakalets, line 8: The column name should be rather `friend_id`

core/src/main/resources/db/changelog/logs/ch-users-friends-Datsko.xml
```diff
@@ -0,0 +1,18 @@
...
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+    </changeSet>
+    <changeSet id="1571227598316-105" author="Marian Datsko">
```
antonbakalets, line 13: If the `users_friends` table supports the many-to-many relation, it should also have a Foreign Key Constraint on `friend_id` column.

PR #885, opened 2020-06-11
Fix eco news dublicated tags and title field limit

TarasBeshlei:
Eco news with duplicated tags will not be created. Title limit changed to 170.


core/src/main/java/greencity/service/impl/EcoNewsServiceImpl.java
```diff
@@ -76,6 +79,13 @@ public AddEcoNewsDtoResponse save(AddEcoNewsDtoRequest addEcoNewsDtoRequest,
         if (image != null) {
             toSave.setImagePath(fileService.upload(image).toString());
         }
+
+        Set<String> tagsSet = new HashSet<>(addEcoNewsDtoRequest.getTags());
+
```
antonbakalets, line 84: Could you please explain this if condition?

- `tagsSet `is created as a copy of `addEcoNewsDtoRequest.getTags()`
- so the size of `tagsSet `is the same as the size of `addEcoNewsDtoRequest.getTags()`
- so the if condition is always false
- so we have an unreachable statement?

TarasBeshlei: addEcoNewsDtoRequest.getTags() this is List which can contain duplicates of tags that not allowed. So I created a Set which will have smaller size than  addEcoNewsDtoRequest.getTags() if it contains duplicates

antonbakalets: Does it makes sense to save the Set without duplicates and do no return the error to user?  To make the correction for user on our own?

TarasBeshlei: In if condition I throw exception, so user will be informed about error. Also I thought about auto correction, but I made it as "Expected result" was required.

antonbakalets: Could you please give me a link to find the "Expected result"?

TarasBeshlei: https://github.com/ita-social-projects/GreenCity/issues/718

PR #897, opened 2020-06-19
Added EcoNews Comment functionality

WayUp06:


core/src/main/java/greencity/controller/EcoNewsCommentController.java
```diff
@@ -0,0 +1,224 @@
...
+     * @return dto {@link greencity.dto.econewscomment.AddEcoNewsCommentDtoResponse}
+     */
+    @ApiOperation(value = "Add comment.")
+    @ApiResponses(value = {
+        @ApiResponse(code = 201, message = HttpStatuses.CREATED, response = CommentReturnDto.class),
+        @ApiResponse(code = 303, message = HttpStatuses.SEE_OTHER),
+        @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
+    })
+    @PostMapping("{econewsId}")
+    public ResponseEntity<Object> save(@PathVariable Long econewsId,
```
antonbakalets, line 58: - Why `Object` is used as generic type?
- Why method is called `save`, javaDoc mentions that it should create smth, the response code is CREATED, but all the method really does is `findByEmail`?

core/src/main/java/greencity/controller/EcoNewsCommentController.java
```diff
@@ -0,0 +1,224 @@
...
+        @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
+    })
+    @GetMapping("")
+    @ApiPageable
+    public ResponseEntity<PageableDto<EcoNewsCommentDto>> findAll(@ApiIgnore Pageable pageable,
+                                                                  Long ecoNewsId,
+                                                                  @ApiIgnore @AuthenticationPrincipal
+                                                                      Principal principal) {
+        User user = null;
+        if (principal != null) {
```
antonbakalets, line 87: Is is possible that principal is null? (I have heard this question one of Academy demos)

core/src/main/java/greencity/controller/EcoNewsCommentController.java
WayUp06: > * Why `Object` is used as generic type?
 It will be fixed in next pull requests
> * Why method is called `save`, javaDoc mentions that it should create smth, the response code is CREATED, but all the method really does is `findByEmail`?
 The main logic of method is located in return statement : .body(ecoNewsCommentService.save(econewsId, addEcoNewsCommentDtoRequest, user))

core/src/main/java/greencity/controller/EcoNewsCommentController.java
WayUp06: > Is is possible that principal is null? (I have heard this question one of Academy demos)
 Yes. If user is not authenticated (didn't sign in), principal is null

PR #899, opened 2020-06-21
Implement facebook service auth

zendermean:
1. Added depencendices.
2. Configure properties file.
3. Refactored SecurityConfig.
4. Added FacebookSecurityControllerController.
5. Added FacebookSecurityService & FacebookSecurityServiceImpl.


core/src/main/java/greencity/config/SecurityConfig.java
```diff
@@ -29,9 +27,7 @@
 import org.springframework.web.cors.CorsConfigurationSource;
 import org.springframework.web.cors.UrlBasedCorsConfigurationSource;
 
-import static greencity.constant.AppConstant.ADMIN;
-import static greencity.constant.AppConstant.MODERATOR;
-import static greencity.constant.AppConstant.USER;
+import static greencity.constant.AppConstant.*;
```
hola-holo, line 30: our static analysis engine detects that your pull request includes a change from an explicit import to a wildcard import. According to the best practice (http://www.javadude.com/articles/importondemandisevil.html), import star (import on demand) is a bad practice, wildcard import can sometimes cause name conflicts and ambiguities.

PR #902, opened 2020-06-23
Fix news in the news recommendations widget

yuriizhurakovskyi:


core/src/main/java/greencity/service/impl/EcoNewsServiceImpl.java
```diff
@@ -119,9 +121,46 @@ public AddEcoNewsDtoResponse save(AddEcoNewsDtoRequest addEcoNewsDtoRequest,
...
+
+    /**
+     * {@inheritDoc}
+     *
+     * @author Zhurakovskyi Yurii.
+     */
+    @Override
+    public List<EcoNewsDto> getThreeRecommendedEcoNewsByTags(List<String> tags, Long openedEcoNewsId) {
+        List<EcoNews> ecoNewsList = ecoNewsRepo.getThreeRecommendedEcoNewsByTags(tags, openedEcoNewsId);
+        if (ecoNewsList.size() < 3) {
```
antonbakalets, line 154: Magic Number Anti-pattern.

core/src/main/java/greencity/controller/EcoNewsController.java
```diff
@@ -165,4 +165,30 @@ public EcoNewsController(EcoNewsService ecoNewsService) {
...
+            @ApiResponse(code = 400, message = HttpStatuses.BAD_REQUEST),
+            @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
+    })
+    @GetMapping("/recommended")
+    public ResponseEntity<List<EcoNewsDto>> getThreeRecommendedEcoNews(
+            @ApiParam(value = "Tags for priorities(if do not input tags get without priorities)")
+            @RequestParam(required = false) List<String> tags,
+            @RequestParam(required = true) Long openedEcoNewsId
+    ) {
+        if (tags == null || tags.isEmpty()) {
```
antonbakalets, line 187: - I would move this check to service level
- I would add 'else' to explicitly indicate that those are two exclusive branches of method execution

core/src/main/java/greencity/service/impl/EcoNewsServiceImpl.java
yuriizhurakovskyi: Replaced by a NUMBER_OF_RECOMMENDED_ECO_NEWS constant

core/src/main/java/greencity/controller/EcoNewsController.java
yuriizhurakovskyi: I combined getThreeRecommendedEcoNews, getThreeRecommendedEcoNewsByTags service methods into a single getThreeRecommendedEcoNews service method, moved tag count check to service level

PR #923, opened 2020-06-29
Refactored econews comment

WayUp06:


core/src/main/java/greencity/entity/enums/CommentStatus.java
```diff
@@ -0,0 +1,7 @@
+package greencity.entity.enums;
+
+public enum CommentStatus {
+    Original,
```
antonbakalets, line 4: Enum names must be in upper case.

core/src/main/java/greencity/service/impl/EcoNewsCommentServiceImpl.java
```diff
@@ -68,7 +68,8 @@ public AddEcoNewsCommentDtoResponse save(Long econewsId, AddEcoNewsCommentDtoReq
      */
     @Override
     public PageableDto<EcoNewsCommentDto> findAllComments(Pageable pageable, User user, Long ecoNewsId) {
-        Page<EcoNewsComment> pages = ecoNewsCommentRepo.findAllByParentCommentIsNullAndEcoNewsIdOrderByCreatedDateDesc(
+        ecoNewsService.findById(ecoNewsId);
```
antonbakalets, line 71: Why does `findById` is called and the result of the call is ignored?

core/src/main/java/greencity/mapping/EcoNewsCommentDtoMapper.java
```diff
@@ -3,24 +3,35 @@
...
-            .build();
+        EcoNewsCommentDto dto = new EcoNewsCommentDto();
+        dto.setId(ecoNewsComment.getId());
+        dto.setModifiedDate(ecoNewsComment.getModifiedDate());
+        if (ecoNewsComment.isDeleted()) {
+            dto.setStatus(CommentStatus.Deleted);
+            return dto;
+        }
+
+        if (ecoNewsComment.getCreatedDate().isEqual(ecoNewsComment.getModifiedDate())) {
```
antonbakalets, line 22: Do I understand correctly that `CommentStatus.Deleted` will be always overridden by `CommentStatus.Original` or `CommentStatus.Edited`?

core/src/main/java/greencity/service/impl/EcoNewsCommentServiceImpl.java
WayUp06: It only checks whether econews with this id exists

PR #930, opened 2020-06-30
Moved TagsController findAllTags method into EcoNewsController

yuriizhurakovskyi:


core/src/main/java/greencity/controller/EcoNewsController.java
```diff
@@ -187,4 +190,15 @@ public EcoNewsController(EcoNewsService ecoNewsService) {
         List<EcoNewsDto> threeRecommendedEcoNews = ecoNewsService.getThreeRecommendedEcoNews(tags, openedEcoNewsId);
         return ResponseEntity.status(HttpStatus.OK).body(threeRecommendedEcoNews);
     }
+
+    /**
+     * The method which returns all tags.
+     *
```
antonbakalets, line 196: Actually it returns not all tags, but the ones attached to eco news. Naming is confusing.

yuriizhurakovskyi: Changed to 'The method which returns all eco news tags.'

PR #954, opened 2020-07-09
Fix API bugs

MarDatsko:


core/src/main/java/greencity/service/impl/UserServiceImpl.java
```diff
@@ -621,6 +621,9 @@ public User updateUserProfilePicture(MultipartFile image, String email) {
         User user = userRepo
             .findByEmail(email)
             .orElseThrow(() -> new WrongEmailException(USER_NOT_FOUND_BY_EMAIL + email));
+        if (image == null) {
```
antonbakalets, line 624: How the image can be null here if you added ImageValidation at controller level? Is ImageValidation working?

core/src/main/java/greencity/validator/ImageValidator.java
```diff
@@ -0,0 +1,25 @@
...
+public class ImageValidator implements ConstraintValidator<ImageValidation, MultipartFile> {
+    private final List<String> validType = Arrays.asList("image/jpeg", "image/png", "image/jpg");
+
+    @Override
+    public void initialize(ImageValidation constraintAnnotation) {
+    }
+
+    @Override
+    public boolean isValid(MultipartFile image, ConstraintValidatorContext constraintValidatorContext) {
+        if (image == null) {
```
antonbakalets, line 19: Now I see. Why null image is valid?

PR #956, opened 2020-07-09
Refactored Habit Statistic Controller Tests

greatpanda1603:


core/src/test/java/greencity/controller/HabitStatisticControllerTest.java
```diff
@@ -1,111 +1,114 @@
...
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.patch;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
-@SpringBootTest
-@AutoConfigureMockMvc
-@RunWith(SpringRunner.class)
+@ExtendWith(SpringExtension.class)
```
antonbakalets, line 31: `@ExtendWith(MockitoExtension.class)` should be enough as there is no Spring context here, only mocked dependencies.

core/src/test/java/greencity/controller/HabitStatisticControllerTest.java
```diff
@@ -1,111 +1,114 @@
...
+    @InjectMocks
+    private HabitStatisticController habitStatisticController;
+
+    @Mock
+    private HabitStatisticService habitStatisticService;
+
+    @BeforeEach
+    void setup() {
+        this.mockMvc = MockMvcBuilders.standaloneSetup(habitStatisticController)
+            .alwaysDo(print())
```
antonbakalets, line 45: Should not be committed.

core/src/test/java/greencity/controller/HabitStatisticControllerTest.java
greatpanda1603: Used Mockito extension, updated respective dependencies in POM

core/src/test/java/greencity/controller/HabitStatisticControllerTest.java
greatpanda1603: Removed .alwaysDo(print())

PR #979, opened 2020-07-22
Added checking by id if a user is online

yuriizhurakovskyi:


core/src/main/java/greencity/repository/UserRepo.java
```diff
@@ -129,4 +129,15 @@
...
     void updateUserLastActivityTime(Long userId, Date userLastActivityTime);
+
+    /**
+     * Find the last activity time by {@link User}'s id.
+     *
+     * @param userId - {@link User}'s id
+     * @return {@link Date}
+     * @author Yurii Zhurakovskyi
+     */
+    @Query(nativeQuery = true,
```
antonbakalets, line 140: You can avoid @Query here and let spring-data handle it for you based on method name convention.

core/src/main/java/greencity/service/impl/UserServiceImpl.java
```diff
@@ -722,4 +726,25 @@ public void addNewFriend(Long userId, Long friendId) {
...
+     * @return {@link Boolean}.
+     * @author Yurii Zhurakovskyi
+     */
+    @Override
+    public Boolean checkIfTheUserIsOnline(Long userId) {
+        Date userLastActivityTime = userRepo.findLastActivityTimeById(userId)
+                .orElseThrow(() -> new UserLastActivityTimeNotFoundException(
+                        USER_LAST_ACTIVITY_TIME_NOT_FOUND + userId));
+        Date currentTime = new Date();
+        long result = currentTime.getTime() - userLastActivityTime.getTime();
```
antonbakalets, line 743: Here is one more reason to have a separate table with not nullable column - this calculation should be part of the query.

core/src/main/java/greencity/service/impl/UserServiceImpl.java
```diff
@@ -722,4 +726,25 @@ public void addNewFriend(Long userId, Long friendId) {
...
+        boolean check = false;
+        if (result <= Long.parseLong(timeAfterLastActivity)) {
```
antonbakalets, line 745: `return result <= Long.parseLong(timeAfterLastActivity);`

core/src/main/java/greencity/service/impl/UserServiceImpl.java
```diff
@@ -57,6 +59,8 @@
     private final HabitDictionaryService habitDictionaryService;
     private final HabitDictionaryTranslationRepo habitDictionaryTranslationRepo;
     private final FileService fileService;
+    @Value("${greencity.time.after.last.activity}")
+    private String timeAfterLastActivity;
```
antonbakalets, line 63: Why `String`, not `long`?

PR #983, opened 2020-07-23
Merge tags

greatpanda1603:
Merge EcoNews Tags and Tips & Tricks Tags to be represented by single entity Tag:
* make necessary adjustments on DB level;
* amend queries to repo ;
* refactor respective services, controllers and tests.


core/src/main/java/greencity/service/impl/EcoNewsServiceImpl.java
```diff
@@ -180,6 +180,7 @@ public AddEcoNewsDtoResponse save(AddEcoNewsDtoRequest addEcoNewsDtoRequest,
      */
     @Override
     public PageableDto<EcoNewsDto> find(Pageable page, List<String> tags) {
+        tags.replaceAll(String::toLowerCase);
```
antonbakalets, line 183: It's a bad style to modify input parameter.

PR #1024, opened 2020-07-29
Added unit tests for AddEcoNewsCommentDtoResponse and Request mappers

yuriizhurakovskyi:


core/src/test/java/greencity/mapping/AddEcoNewsCommentDtoRequestMapperTest.java
```diff
@@ -0,0 +1,27 @@
...
+import greencity.entity.EcoNewsComment;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import static greencity.ModelUtils.getEcoNewsComment;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@ExtendWith(SpringExtension.class)
```
antonbakalets, line 14: You are not using Spring context in this test. `@ExtendWith(MockitoExtension.class)` should be enough.

antonbakalets: Actually `AddEcoNewsCommentDtoRequestMapper` don't have any dependency that must be injected, you can just create an instance of it and run the test without any Extension.

yuriizhurakovskyi: Removed Extensions

PR #1062, opened 2020-07-31
Added unit tests for TipsAndTricksCommentController

yuriizhurakovskyi:


core/src/test/java/greencity/controller/TipsAndTricksCommentControllerTest.java
```diff
@@ -0,0 +1,182 @@
...
+                .principal(principal)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(content))
+                .andExpect(status().isCreated());
+
+        ObjectMapper mapper = new ObjectMapper();
+        AddTipsAndTricksCommentDtoRequest addTipsAndTricksCommentDtoRequest = mapper.readValue(content, AddTipsAndTricksCommentDtoRequest.class);
+
+        verify(userService).findByEmail(eq("Liam.Johnson@gmail.com"));
+        verify(tipsAndTricksCommentService)
```
antonbakalets, line 75: `eq(new User())` is bad. In fact it should be equal not to a new user, but to the user returned by `userService.findByEmail`. Looks like the test pass because `equals` is not implemented.

yuriizhurakovskyi: User class has such annotation:
@EqualsAndHashCode(
    exclude = {"places", "comments", "verifyEmail", "addedPlaces", "favoritePlaces", "ownSecurity", "refreshTokenKey",
        "verifyEmail", "estimates", "restorePasswordEmail", "addedEcoNews", "addedTipsAndTricks"})

Replaced new User() to User from ModelUtils

antonbakalets: Yes, much better now. Looks like before the check `new User().equals(new User())` was returning true, which is not really true.

PR #1113, opened 2020-08-10
Fixed error for user without habits (HabitStatisticServiceImpl)

Yura09:


core/src/test/java/greencity/service/impl/HabitStatisticServiceImplTest.java
```diff
@@ -67,19 +58,18 @@
 
     private Habit habit = new Habit(1L, new HabitDictionary(), new User(), true,
         zonedDateTime, Collections.emptyList());
+    private HabitStatistic habitStatistic = new HabitStatistic(
+        1L, HabitRate.GOOD, ZonedDateTime.now(), 10, null);
 
     @BeforeEach
-    public void init() {
+    void init() {
         MockitoAnnotations.initMocks(this);
```
antonbakalets, line 66: This is what InjectMocs and MockExtention are doing behind the scene.

PR #1127, opened 2020-08-10
Set up thymeleaf, added live-reload

vorfolomey:


core/src/main/java/greencity/config/SwaggerConfig.java
```diff
@@ -56,7 +58,9 @@ public Docket swaggerSpringfoxDocket() {
                 .securitySchemes(Lists.newArrayList(apiKey()))
                 .useDefaultResponseMessages(false);
 
-        docket = docket.select().apis(RequestHandlerSelectors.withClassAnnotation(RestController.class)).build();
+        docket =
+            docket.select().apis(RequestHandlerSelectors.any())
```
antonbakalets, line 62: Can we specify Controller and RestController only instead of any?

vorfolomey: > Can we specify Controller and RestController only instead of any?

I did not find a solution for this

PR #1141, opened 2020-08-11
Fixed bugs in TipsAndTricksCommentServiceImpl

Yura09:


core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -36,6 +36,21 @@
...
+     *
+     * @param ex      Exception witch should be intercepted.
+     * @param request contain  detail about occur exception
+     * @return ResponseEntity witch  contain http status and body  with message of exception.
+     * @author Yurii Savchenko
+     */
+    @ExceptionHandler(BadRequestException.class)
+    public final ResponseEntity<Object> handleBadRequestException(BadRequestException ex, WebRequest request) {
+        ExceptionResponse exceptionResponse = new ExceptionResponse(getErrorAttributes(request));
+        log.trace(ex.getMessage(), ex);
```
antonbakalets, line 50: This needs a little improvment.
1. TRACE level is too low, I would use INFO.
2. The exception is handled, so no need to print the stack trace. Please remove the last parameter.
3. Please move logging to the first line: log the exception before any interaction.

PR #1154, opened 2020-08-13
Fact of the day repos services rest cache

vorfolomey:


core/src/main/resources/application-dev.properties
```diff
@@ -13,7 +13,7 @@ spring.liquibase.enabled=true
 spring.liquibase.change-log=classpath:db/changelog/db.changelog-master.xml
 
 # Hibernate
-spring.jpa.hibernate.ddl-auto=validate
+spring.jpa.hibernate.ddl-auto=update
```
antonbakalets, line 16: Please revert. As far as I know in this project it is the responsibility of liquibase to update the database s hema.

vorfolomey: Okay done.

PR #1164, opened 2020-08-16
Getting User online status for a user and friends using sockets

yuriizhurakovskyi:


core/src/main/java/greencity/service/impl/UserServiceImpl.java
```diff
@@ -775,15 +777,28 @@ public void updateUserLastActivityTime(Long userId, Date userLastActivityTime) {
...
-        Date currentTime = new Date();
-        long result = currentTime.getTime() - userLastActivityTime.getTime();
-        return result <= timeAfterLastActivity;
+    public boolean checkIfTheUserIsOnline(Long userId) {
+        if (!userRepo.findById(userId).isPresent()) {
+            throw new WrongIdException(USER_NOT_FOUND_BY_ID + userId);
+        }
+        try {
+            LocalDateTime userLastActivityTime = userRepo.findLastActivityTimeById(userId)
+                    .orElseThrow(() -> new UserLastActivityTimeNotFoundException(USER_LAST_ACTIVITY_TIME_NOT_FOUND));
```
antonbakalets, line 786: So you are throwing an exception just to catch it few lines later. Why don't you just check `Optional.isPresent()` in an `if` statement

yuriizhurakovskyi: Throwing an UserLastActivityTimeNotFoundException was replaced to checking Optional.isPresent() in checkIfTheUserIsOnline method

antonbakalets (approved):

Yura09 (approved):

eyouu (approved):

PR #1176, opened 2020-08-18
Get all active comments

TarasDovganyuk:
Method getAllActiveComments in EcoNewsCommentController, service and repo added.


core/src/main/java/greencity/service/impl/EcoNewsCommentServiceImpl.java
```diff
@@ -212,4 +213,30 @@ public int countReplies(Long id) {
...
+     *
+     * @param user      current {@link User}
+     * @param pageable  page of news.
+     * @param ecoNewsId specifies {@link greencity.entity.EcoNews} to which we search for comments
+     * @return all active comments to certain ecoNews specified by ecoNewsId.
+     * @author Taras Dovganyuk
+     */
+    @Override
+    public PageableDto<EcoNewsCommentDto> getAllActiveComments(Pageable pageable, User user, Long ecoNewsId) {
+        Page<EcoNewsComment> pages =
```
antonbakalets, line 228: If parameter `user` is not used you should remove it.

antonbakalets (changes requested):

antonbakalets (approved):

PR #1183, opened 2020-08-19
Management "Update" User

eyouu:


core/src/main/java/greencity/dto/user/UserManagementDto.java
```diff
@@ -0,0 +1,36 @@
...
+    private Long id;
+
+    @NotBlank(message = ValidationConstants.EMPTY_USERNAME)
+    @Size(
+        min = ValidationConstants.USERNAME_MIN_LENGTH,
+        max = ValidationConstants.USERNAME_MAX_LENGTH,
+        message = ValidationConstants.INVALID_USERNAME_LENGTH)
+    private String name;
+
+    @Email(message = ValidationConstants.INVALID_EMAIL)
```
antonbakalets, line 28: Those messages are not i18n, but the application is intended to support at least 3 languages. Was the problem with error messages translation solved?

antonbakalets (approved):

TarasDovganyuk (approved):

Yura09 (approved):

core/src/main/java/greencity/dto/user/UserManagementDto.java
eyouu: I did not see any problems with this, and also as far as I remember we decided to do our admin panel only in English. 
image

PR #1233, opened 2020-08-26
Deleting deactivated users functionality

eyouu:
The logic of setting status 'DEACTIVATED' for the user.
Deleting from the database users that have status 'DEACTIVATED' and last visited the site 2 years ago.


MarDatsko (approved):

core/src/main/java/greencity/webcontroller/ManagementUserController.java
```diff
@@ -76,13 +78,27 @@ public String updateUser(@Valid @ModelAttribute("user") UserManagementDto userDt
...
+     * Method for setting {@link User}'s status to DEACTIVATED,
+     * so the user will not be able to log in into the system.
+     *
+     * @param id of the searched {@link User}.
+     * @author Vasyl Zhovnir
+     */
+    @PostMapping
+    public ResponseEntity<UserDeactivateDto> deactivateUser(Long id) {
+        User byId = userService.findById(id);
+        UserDeactivateDto dto = modelMapper.map(byId, UserDeactivateDto.class);
```
antonbakalets, line 101: You are: 

- extracting the user from the database by _id_
- converting to dto and pass it to `userService.deactivateUser(dto)`
- finally ... in method `deactivateUser` get only the _id_ from the dto to find user once again

I think you can pass the _id_ directly to service method `deactivateUser`

antonbakalets (changes requested):

core/src/main/java/greencity/webcontroller/ManagementUserController.java
eyouu: Thank you, done.

antonbakalets (approved):

TarasDovganyuk (approved):

PR #1235, opened 2020-08-26
Changed incorrect message and response code

Yura09:


MarDatsko (approved):

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -36,6 +37,23 @@
 public class CustomExceptionHandler extends ResponseEntityExceptionHandler {
     private ErrorAttributes errorAttributes;
 
+    /**
+     * Method intercept exception {@link UserHasNoAvailableHabitDictionaryException}.
```
antonbakalets, line 41: JavaDoc is mentioning `UserHasNoAvailableHabitDictionaryException`, but the annotation is mentioning `ConstraintViolationException`. So which one are you trying to handle?

core/src/main/java/greencity/repository/HabitDictionaryTranslationRepo.java
```diff
@@ -38,9 +38,11 @@
      * @param language code language.
      * @return List of available {@link HabitDictionaryTranslation}`s.
      */
-    @Query(nativeQuery = true, value = "SELECT *  FROM habit_dictionary_translation WHERE"
-        + " habit_dictionary_id NOT IN\n"
-        + " (SELECT habits.habit_dictionary_id FROM habits WHERE user_id = ?1 AND status = 'true')\n"
+    @Query(nativeQuery = true, value = "SELECT *  FROM habit_dictionary_translation WHERE "
+        + "habit_dictionary_id NOT IN "
```
antonbakalets, line 42: image
Is that what you are trying to achieve with this NOT IN?

antonbakalets (changes requested):

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -36,6 +37,39 @@
...
+    /**
+     * Method intercept exception {@link ConstraintViolationException}.
+     *
+     * @param ex      Exception witch should be intercepted.
+     * @param request contain  detail about occur exception
+     * @return ResponseEntity witch  contain http status and body  with message of exception.
+     * @author Yurii Savchenko
+     */
+    @ExceptionHandler(ConstraintViolationException.class)
+    public final ResponseEntity<Object> handleConstraintViolationException(ConstraintViolationException ex,
```
antonbakalets, line 65: Now there is a new problem. `ConstraintViolationException` is a generic exception that is thrown on any validation error. So whatever is the reason user will receive `SELECT_CORRECT_LANGUAGE` message. 
Can you send instead the message extracted from `ex` object instead? 
Could you please check that my suggestions are correct?

TarasDovganyuk (approved):

PR #1241, opened 2020-08-28
Change the response to return message without the name of the function's invocation

vorfolomey:


core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -66,6 +67,10 @@
                                                                            WebRequest request) {
         log.info(ex.getMessage());
         ExceptionResponse exceptionResponse = new ExceptionResponse(getErrorAttributes(request));
+        String detailedMessage = ex.getConstraintViolations().stream()
+            .map(ConstraintViolation::getMessage)
+            .reduce("", (subTotal, element) -> subTotal + element);
```
antonbakalets, line 72: Using `.reduce()` you concatenate `String`s producing a new `String` each time.
It's better to use `.collect(Collectors.joining(""));` that is designed for such cases.

antonbakalets (changes requested):
I was waiting for this pull request.

MarDatsko (approved):

TarasDovganyuk (approved):

eyouu (approved):

antonbakalets (approved):

PR #1266, opened 2020-09-02
Rating calculation

TarasDovganyuk:
Created aspect where User rating is calculating


Yura09 (approved):

MarDatsko (approved):

antonbakalets:
As discussed:

- lets replace method names with an enum.
- lets receive current user from SecurityContextHolder, not by parsing arguments.
- lets introduce a service that can be reused by admin part.

core/src/main/java/greencity/service/impl/PlaceCommentServiceImpl.java
```diff
@@ -83,8 +86,9 @@ public CommentReturnDto save(Long placeId, AddCommentDto addCommentDto, String e
      *
      * @author Marian Milian
      */
+    @RatingCalculation(rating = RatingCalculationEnum.DELETE_COMMENT)
     @Override
-    public void deleteById(Long id) {
+    public void deleteById(Long id, User user) {
```
antonbakalets, line 91: You don't need this User parameter anymore, did you?

core/src/main/java/greencity/service/impl/PlaceCommentServiceImpl.java
```diff
@@ -93,13 +97,13 @@ public void deleteById(Long id) {
      * {@inheritDoc}
      */
     @Override
-    public PageableDto getAllComments(Pageable pageable) {
+    public PageableDto<CommentAdminDto> getAllComments(Pageable pageable) {
```
antonbakalets, line 100: Like.

core/src/main/java/greencity/annotations/RatingCalculationEnum.java
```diff
@@ -0,0 +1,21 @@
+package greencity.annotations;
+
+import lombok.Getter;
+
+@Getter
+public enum RatingCalculationEnum {
```
antonbakalets, line 6: There is no need to add `Enum` to the name, the same way as we don't add `Class` to the class name or `I` to interface name.

antonbakalets (approved):

PR #1294, opened 2020-09-04
Deauthorize deactivated user

eyouu:


vorfolomey (approved):

MarDatsko (approved):

Yura09 (approved):

core/src/main/java/greencity/security/filters/AccessTokenAuthenticationFilter.java
```diff
@@ -24,35 +28,42 @@
...
         String token = jwtTool.getTokenFromHttpServletRequest(request);
+
         if (token != null) {
             try {
                 Authentication authentication = authenticationManager
                     .authenticate(new UsernamePasswordAuthenticationToken(token, null));
-                log.info("User successfully authenticate - {}", authentication.getPrincipal());
-                SecurityContextHolder.getContext().setAuthentication(authentication);
+                User user = userService.findByEmail((String) authentication.getPrincipal());
+                if (user.getUserStatus() != UserStatus.DEACTIVATED) {
```
antonbakalets, line 63: My first idea was that this new if statement is not covered by unit test. Have you measured the test coverage for this class?

antonbakalets: But then I've got a better idea. You don't have to load the user and then check the status. You should create a new method in `UserService` that will query the database for a user "by email and status equals not deactivated".

antonbakalets: Another problem with the current approach is that if user is not found, a NPE will be thrown.

antonbakalets (changes requested):

core/src/main/java/greencity/security/filters/AccessTokenAuthenticationFilter.java
eyouu: Done!

antonbakalets (approved):

PR #1297, opened 2020-09-04
Management. Popups for user

eyouu:


Yura09 (approved):

core/src/main/java/greencity/service/impl/UserServiceImpl.java
```diff
@@ -925,9 +925,30 @@ public UserAndAllFriendsWithOnlineStatusDto getAllFriendsWithTheOnlineStatus(Lon
...
         foundUser.setUserStatus(UserStatus.DEACTIVATED);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Transactional
+    @Override
+    public List<Long> deactivateAllUsers(List<Long> listId) {
+        listId.forEach(this::deactivateUser);
```
antonbakalets, line 941: There is a problem with this approach. What if in the middle of the list a user is not found? It will fail in the middle, does it mean that some users from the list will be deactivated and some not?

core/src/main/java/greencity/webcontroller/ManagementUserController.java
```diff
@@ -64,13 +55,19 @@ public String showForm(@RequestParam Long id, Model model) {
...
+            GenericResponseDto genericResponseDto = new GenericResponseDto();
+            for (FieldError fieldError : bindingResult.getFieldErrors()) {
+                genericResponseDto.getErrors().add(
+                    new FieldErrorDto(fieldError.getField(), fieldError.getDefaultMessage()));
+            }
+            return genericResponseDto;
         }
         userService.updateUser(userDto);
-        return "redirect:/management/users";
+        return GenericResponseDto.builder().errors(new ArrayList<>()).build();
```
antonbakalets, line 70: Is it possible to remove `.errors(new ArrayList<>())`. For example make it the builder responsibility to initialize the list as empty?

antonbakalets (changes requested):

core/src/main/java/greencity/webcontroller/ManagementUserController.java
eyouu: Yes, done.

core/src/main/java/greencity/service/impl/UserServiceImpl.java
eyouu: Changed approach.

antonbakalets (approved):

TarasDovganyuk (approved):

PR #1301, opened 2020-09-04
Service for custom social network icons

vorfolomey:


core/src/main/java/greencity/dto/user/UserProfileDtoResponse.java
```diff
@@ -14,7 +16,7 @@
     private String firstName;
     private String city;
     private String userCredo;
-    private List<String> socialNetworks;
+    private List<SocialNetworkResponseDTO> socialNetworks = new ArrayList<>();
```
antonbakalets, line 19: Field initialization is the Mapper responsibility.

core/src/main/java/greencity/mapping/MultipartBase64ImageMapper.java
```diff
@@ -40,7 +37,14 @@ protected MultipartFile convert(String image) {
             ImageIO.write(bufferedImage, "png", tempFile);
             FileItem fileItem = new DiskFileItem("mainFile", Files.probeContentType(tempFile.toPath()),
                 false, tempFile.getName(), (int) tempFile.length(), tempFile.getParentFile());
-            fileItem.getOutputStream();
+            InputStream input = new FileInputStream(tempFile);
```
antonbakalets, line 40: - This is not really optimal to copy byte by byte.
- IOStreams should be closed (using try-with-resources block)

To sum up: you can use https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/IOUtils.html#copy-java.io.InputStream-java.io.OutputStream-

Here and in two more places

core/src/main/java/greencity/entity/SocialNetwork.java
```diff
@@ -0,0 +1,28 @@
...
+public class SocialNetwork {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @Length(min = 1, max = 500)
+    @Column(name = "social_network_url", length = 500)
+    String url;
+
+    @ManyToOne(fetch = FetchType.LAZY)
```
antonbakalets, line 23: I don't think it should be LAZY from this side of the connection.

core/src/main/java/greencity/entity/SocialNetwork.java
```diff
@@ -0,0 +1,28 @@
...
+    User user;
+
+    @OneToOne(fetch = FetchType.LAZY)
+    SocialNetworkImage socialNetworkImage;
```
antonbakalets, line 27: This one also not really LAZY as almost always you will be loading user social networks with corresponding image.

core/src/main/java/greencity/service/impl/SocialNetworkServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        List<SocialNetwork> listSocialNetwork = socialNetworkUrls.stream()
+            .map(url ->
+                SocialNetwork.builder()
+                    .url(url)
+                    .user(user)
+                    .socialNetworkImage(socialNetworkImageService.getSocialNetworkImageByUrl(url))
+                    .build()
+            )
+            .collect(Collectors.toList());
+        socialNetworkRepo.deleteAllByUserIs(user);
```
antonbakalets, line 39: Can you please check if the following will work instead (please check logged SQL statements):
```
user.getSocialNetworks().clear();
user.setSocialNetworks(listSocialNetwork);
```
If that is the case you will be able to remove those Repos. Sometime JPA can do some magic...

antonbakalets (changes requested):

core/src/main/java/greencity/entity/SocialNetwork.java
vorfolomey: What does that mean?

core/src/main/java/greencity/entity/SocialNetwork.java
vorfolomey: Okay

Yura09 (approved):

core/src/main/java/greencity/service/impl/SocialNetworkServiceImpl.java
vorfolomey: Попробував. Не працює чогось. Тільки в об'єкті зміни, але в бд нічого не міняється

TarasDovganyuk (approved):

MarDatsko (approved):

eyouu (approved):

core/src/main/java/greencity/service/impl/SocialNetworkServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        List<SocialNetwork> listSocialNetwork = socialNetworkUrls.stream()
+            .map(url ->
+                SocialNetwork.builder()
+                    .url(url)
+                    .user(user)
+                    .socialNetworkImage(socialNetworkImageService.getSocialNetworkImageByUrl(url))
+                    .build()
+            )
+            .collect(Collectors.toList());
+        user.getSocialNetworks().clear();
```
antonbakalets, line 37: Please move those two lines out of this method (one level up), then:

- remove parameter user

antonbakalets (approved):

PR #1329, opened 2020-09-09
Searching methods for EcoNews and Tips&Tricks

Yura09:


core/src/main/resources/db/functions/vw_econewswithtags.sql
```diff
@@ -0,0 +1,16 @@
+CREATE MATERIALIZED VIEW public.vw_econewswithtags
```
antonbakalets, line 1: Please add IF NOT EXISTS or the view will be recreated on each application startup.

antonbakalets (approved):

core/src/main/resources/db/functions/vw_econewswithtags.sql
Yura09: ok)

forestik (approved):

vorfolomey (approved):

TarasDovganyuk (approved):

PR #1353, opened 2020-09-10
Unit test for FactOfTheDayService

SashkoMolodec:


core/src/test/java/greencity/service/impl/FactOfTheDayServiceImplTest.java
```diff
@@ -0,0 +1,174 @@
...
+import static org.mockito.Mockito.*;
+import org.mockito.MockitoAnnotations;
+import org.modelmapper.ModelMapper;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageImpl;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+@ExtendWith(SpringExtension.class)
```
antonbakalets, line 33: You don't need SpringExtension as spring context is not used in the test, MockitoExtension would be enough.

core/src/test/java/greencity/service/impl/FactOfTheDayServiceImplTest.java
```diff
@@ -0,0 +1,174 @@
...
+public class FactOfTheDayServiceImplTest {
+    @InjectMocks
+    private FactOfTheDayServiceImpl factOfTheDayService;
```
antonbakalets, line 36: The field annotated with `@InjectMocks` should be the last one. When I read the code, I want to see all the `@Mock`-ed objects and then at the end the object that those mocks are injected in.

core/src/test/java/greencity/service/impl/FactOfTheDayServiceImplTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Mock
+    private FactOfTheDayRepo factOfTheDayRepo;
+
+    @Mock
+    private LanguageService languageService;
+
+    @BeforeEach
+    public void setup(){
+        MockitoAnnotations.initMocks(this);
```
antonbakalets, line 52: You don't need this as both `SpringExtension` and `MockitoExtension `are doing this for you under the hood.

antonbakalets (changes requested):

core/src/test/java/greencity/service/impl/FactOfTheDayServiceImplTest.java
SashkoMolodec: Знімок екрана 2020-09-11 084226

SashkoMolodec: For some reason it doesn't

SashkoMolodec: Objects that aren't calling directly in test class are not getting mocked

core/src/test/java/greencity/service/impl/FactOfTheDayServiceImplTest.java
SashkoMolodec: Changes done.

core/src/test/java/greencity/service/impl/FactOfTheDayServiceImplTest.java
SashkoMolodec: Changes done.

core/src/test/java/greencity/service/impl/FactOfTheDayServiceImplTest.java
antonbakalets: I've found the problem. Part of fields in `FactOfTheDayServiceImpl` are initialized via constructor, another part via `@Autovired` annotation. `Mockito` cannot handle both. Need a consistent approach. I suggest, `FactOfTheDayServiceImpl` constructor should be removed.

SashkoMolodec: So should I close this request change? Or make changes in FactOfTheDayServiceImpl?

antonbakalets: Writing unit tests sometime requires target class refactoring. Especially, when the unit test discovers a problem in the target class. Please update `FactOfTheDayServiceImpl`.

SashkoMolodec: Changes done.

SashkoMolodec:
image

PR #1362, opened 2020-09-11
fixed sonar bug

eyouu:


core/src/main/java/greencity/service/impl/CloudStorageService.java
```diff
@@ -82,12 +82,11 @@ public MultipartFile convertToMultipartImage(String image) {
...
         byte[] imageByte = decodeBase64(imageToConvert);
         ByteArrayInputStream bis = new ByteArrayInputStream(imageByte);
-        try {
+        try (InputStream input = new FileInputStream(tempFile)) {
             BufferedImage bufferedImage = ImageIO.read(bis);
             ImageIO.write(bufferedImage, "png", tempFile);
             FileItem fileItem = new DiskFileItem("mainFile", Files.probeContentType(tempFile.toPath()),
                 false, tempFile.getName(), (int) tempFile.length(), tempFile.getParentFile());
-            InputStream input = new FileInputStream(tempFile);
             OutputStream outputStream = fileItem.getOutputStream();
```
antonbakalets, line 90: `OutputStream` is also `AutoCloasable`, so it should be also enclosed into an try-with-resources block.

antonbakalets: Actually there is a better solution - to merge this method with `greencity.mapping.MultipartBase64ImageMapper#convert`

But the best solution would be to inject `greencity.mapping.MultipartBase64ImageMapper` bean (perfectly referenced by interface `org.modelmapper.Converter`) and call its method instead of this copy-paste.

Note usage of `IOUtils.copy(input, outputStream);` instread of this `while`.

antonbakalets (changes requested):

core/src/main/java/greencity/service/impl/CloudStorageService.java
eyouu: It wasn't possible to inject `greencity.mapping.MultipartBase64ImageMapper` referenced by interface `org.modelmapper.Converter` because our `MultipartBase64ImageMapper` overrides own method of `abstact class AbstractConverter`. And I also wasn't able to inject by this `abstact class AbstractConverter` reference because method `convert` in this class is protected.

vorfolomey (approved):

MarDatsko (approved):

antonbakalets (approved):

olenapetryshak (approved):

Issue #1271, opened 2020-09-02
[API] [Edit profile] user can add the same links on social networks


stasgoodman:
**Environment**: Windows 10 Home, GoogleChrome 85.0.4183.83 32-bit
**Reproducible**: always
**Build found:** : commit 4a8f0b6

**Preconditions**
1. Open Swagger: https://greencity.azurewebsites.net/swagger-ui.html#/ ;
2. Sign in with controller Post /ownSecurity/signIn (e.g. green.city.test2@gmail.com, Green.city#500);
3. Use token to authorize in Swagger

**Steps to reproduce**
1. Click on  'user-controller' dropdown 
2. Click on 'Put /user/profile Save user profile information' request
3. Click 'Try it out' and use "{   "city": "string",   "firstName": "string",   "showEcoPlace": true,   "showLocation": true,   "showShoppingList": true,   "socialNetworks": [     "https:// facebook.com/greencity1", "https://facebook.com/greencity1"   ],   "userCredo": "string" }"
4. Click 'Execute' button

**Actual result**
Get the responce 201. Information is saved.

**Expected result**
Should be the responce 400 - user cannot add the same links.

**User story and test case links**
E.g.: "User story #1007 
[Test case] https://jira.softserve.academy/browse/GC-

PR #1364, opened 2020-09-11
Fixed issue with validation SocialNetwork

vorfolomey:
#1271


core/src/main/java/greencity/validator/SocialNetworkLinksValidator.java
```diff
@@ -0,0 +1,36 @@
...
+        }
+        if (!areDistinct(links)) {
+            throw new BadSocialNetworkLinksException(ErrorMessage.USER_CANNOT_ADD_SAME_SOCIAL_NETWORK_LINKS);
+        }
+        links.forEach(UrlValidator::isUrlValid);
+        return true;
+    }
+
+    private boolean areDistinct(List<String> list) {
+        Set<String> hashSet = new HashSet<>(list);
```
antonbakalets, line 33: I'm not sure it is a good idea to check if links are distinct. You can convert behind the scene the `List` to a `Set` so they are distinct and do not borrow user to search which link has a duplicate. Then check `links.size() > ValidationConstants.MAX_AMOUNT_OF_SOCIAL_NETWORK_LINKS`. Please review requirements once again.

antonbakalets (approved):

TarasDovganyuk (approved):

olenapetryshak (approved):

core/src/main/java/greencity/validator/SocialNetworkLinksValidator.java
vorfolomey: image

vorfolomey: What I can do about this?

PR #1365, opened 2020-09-12
Management security

eyouu:
Enum `ROLE` will be replaced with type `String` in the next PR


core/src/main/java/greencity/security/filters/AccessTokenAuthenticationFilter.java
```diff
@@ -55,9 +58,13 @@ public void doFilterInternal(@SuppressWarnings("NullableProblems") HttpServletRe
         String token = jwtTool.getTokenFromHttpServletRequest(request);
         if (token != null) {
             try {
+                Optional<User> user = userService.findNotDeactivatedByEmail(jwtTool.getEmailOutOfAccessToken(token));
+                Collection<? extends GrantedAuthority> authorities = null;
+                if (user.isPresent()) {
```
antonbakalets, line 63: Please move this if inside `convertRoles` method.

core/src/test/java/greencity/security/filters/AccessTokenAuthenticationFilterTest.java
```diff
@@ -18,6 +18,7 @@
 import org.powermock.modules.junit4.PowerMockRunner;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.userdetails.UserDetailsService;
```
antonbakalets, line 21: Please add a unit test covering the case when user has a role.

antonbakalets (changes requested):

core/src/main/java/greencity/security/filters/AccessTokenAuthenticationFilter.java
eyouu: Did you mean that I have to do it the way I did? Since Google says that passing an object enclosed in an Optional into a method is considered bad practice.

core/src/test/java/greencity/security/filters/AccessTokenAuthenticationFilterTest.java
```diff
@@ -44,13 +50,17 @@ public void setUp() {
 
     @Test
     public void doFilterInternal() throws IOException, ServletException {
+        Optional<User> user = Optional.of(ModelUtils.getUser());
+        List<GrantedAuthority> authorities = List.of(new SimpleGrantedAuthority("ROLE_USER"));
```
antonbakalets, line 54: I was expecting to see 

- `greencity.entity.enums.ROLE` as input parameter
- `List<GrantedAuthority>` as output parameter checked somewhere in assertions

to make sure that the conversion was correct.

core/src/main/java/greencity/security/filters/AccessTokenAuthenticationFilter.java
```diff
@@ -70,4 +74,15 @@ public void doFilterInternal(@SuppressWarnings("NullableProblems") HttpServletRe
...
     }
+
+    /**
+     * Convert comma separated string of {@link User} roles to collection of roles.
+     *
+     * @param user - {@link Optional} of {@link User}
+     * @return {@link Collection} of roles
+     */
+    private Collection<? extends GrantedAuthority> convertRoles(Optional<User> user) {
+        return user.map(value -> AuthorityUtils.commaSeparatedStringToAuthorityList(value.getRole().name()))
```
antonbakalets, line 85: Yes, this is what I meant. Taking into account that now it is one line and that passing an Optional is considered a bad practice, you can just inline this method.

PR #1421, opened 2020-09-18
Make return same fact for 3 different languages #1419

vorfolomey:
#1419

### DIFF
--- a/core/src/main/java/greencity/service/impl/FactOfTheDayServiceImpl.java
+++ b/core/src/main/java/greencity/service/impl/FactOfTheDayServiceImpl.java
@@ -1,9 +1,11 @@
 package greencity.service.impl;
 
+import greencity.constant.CacheConstants;
 import greencity.constant.ErrorMessage;
 import greencity.dto.PageableDto;
 import greencity.dto.factoftheday.FactOfTheDayDTO;
 import greencity.dto.factoftheday.FactOfTheDayPostDTO;
+import greencity.dto.factoftheday.FactOfTheDayTranslationDTO;
 import greencity.entity.FactOfTheDay;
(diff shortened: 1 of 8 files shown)


core/src/main/java/greencity/service/impl/FactOfTheDayServiceImpl.java
```diff
@@ -27,14 +34,17 @@
...
+@Service
+@Scope(proxyMode = ScopedProxyMode.TARGET_CLASS)
 public class FactOfTheDayServiceImpl implements FactOfTheDayService {
     private final FactOfTheDayRepo factOfTheDayRepo;
     private final ModelMapper modelMapper;
     private final LanguageService languageService;
     private final FactOfTheDayTranslationService factOfTheDayTranslationService;
     private final PlaceCommentServiceImpl placeCommentServiceImpl;
+    @Resource
+    private FactOfTheDayService self;
```
antonbakalets, line 47: What is the problem you are trying to solve here? Why not use `this` instead?

vorfolomey: https://stackoverflow.com/questions/16899604/spring-cache-cacheable-not-working-while-calling-from-another-method-of-the-s

vorfolomey: The cache doesn`t work from a method inside the same class.

vorfolomey: image

antonbakalets (approved):

forestik (approved):

olenapetryshak (approved):

Yura09 (approved):

PR #1423, opened 2020-09-18
Delete created user

eyouu:


vorfolomey (approved):

core/src/main/java/greencity/config/ScheduleConfig.java
```diff
@@ -134,4 +134,16 @@ public void chooseNewFactOfTheDay() {
     public void scheduleDeleteDeactivatedUsers() {
         userRepo.scheduleDeleteDeactivatedUsers();
     }
+
+    /**
+     * Every day at 00:00 deletes from the database users
+     * that have status 'CREATED' and have not activated the account within 24 hours.
```
antonbakalets, line 140: So If I start my registration at 23:55, I won't be able to finish it at 00:00.

I believe you should delete those, whose creation date is before yesterday.

core/src/main/java/greencity/repository/UserRepo.java
```diff
@@ -196,4 +196,15 @@
...
+
+    /**
+     * Delete from the database users that have status 'CREATED'
+     * and have not activated the account within 24 hours.
+     *
+     * @return number of deleted rows
+     * @author Vasyl Zhovnir
+     **/
+    @Modifying
+    @Query("DELETE FROM User u WHERE u.userStatus = 3 AND u.id NOT IN (SELECT r.user.id FROM RestorePasswordEmail r)")
```
antonbakalets, line 208: Must be a JOIN.

antonbakalets (changes requested):

core/src/main/java/greencity/config/ScheduleConfig.java
eyouu: The idea was the following:
If we create the User (gets `CREATED` status) then the appropriate record will be created in `restore_password_email`  (records in this table are automatically deleted after 86400000 seconds = 24 hours after they were created). So I was checking to see if we have a record in this table for our user. If there were no records - mean 24 hours have passed and we can delete this user.
***
But for now, I redid it to check `date_of_registration` as you suggested.

Yura09 (approved):

TarasDovganyuk (approved):

core/src/main/java/greencity/config/ScheduleConfig.java
antonbakalets: When does the `restore_password_email` table is cleaned?

eyouu: image

antonbakalets: Shouldn't this be triggered together with `userRepo.scheduleDeleteCreatedUsers()` ?

antonbakalets (approved):

core/src/main/java/greencity/config/ScheduleConfig.java
antonbakalets: Ah, that is different. Never mind.

PR #1453, opened 2020-09-22
changed eco-news-comment-controller, fixed search path in security

Yura09:


eyouu (approved):

forestik (approved):

core/src/main/java/greencity/controller/EcoNewsCommentController.java
```diff
@@ -109,7 +123,12 @@ public int getCountOfComments(@PathVariable Long ecoNewsId) {
     @ApiPageable
     public ResponseEntity<PageableDto<EcoNewsCommentDto>> findAllReplies(@ApiIgnore Pageable pageable,
                                                                          @PathVariable Long parentCommentId,
-                                                                         @ApiIgnore @CurrentUser User user) {
+                                                                         @ApiIgnore @AuthenticationPrincipal
+                                                                             Principal principal) {
+        User user = null;
+        if (principal != null) {
```
antonbakalets, line 129: That is too much or revert. May be it then make sense to allow `greencity.converters.UserArgumentResolver#resolveArgument` return null if Principal is not available.

core/src/main/java/greencity/converters/UserArgumentResolver.java
```diff
@@ -34,14 +31,13 @@ public boolean supportsParameter(MethodParameter parameter) {
...
      * @return {@link User}
      */
     @Override
     public Object resolveArgument(MethodParameter parameter, ModelAndViewContainer mavContainer,
                                   NativeWebRequest webRequest, WebDataBinderFactory binderFactory) {
         Principal principal = webRequest.getUserPrincipal();
-        return Optional.ofNullable(userService.findByEmail(Objects.requireNonNull(principal).getName()))
-            .orElseThrow(() -> new WrongEmailException(ErrorMessage.USER_NOT_FOUND_BY_EMAIL + principal.getName())
-            );
+        return userService.findByEmail(principal != null ? principal.getName() : null);
```
antonbakalets, line 41: Please do not search by `null`. There is no reason to call `findByEmail` with argument equal to `null` if you know the result in advance.

antonbakalets (changes requested):

PR #1467, opened 2020-09-23
changed rights for getting comments and replies

Yura09:


core/src/main/java/greencity/config/SecurityConfig.java
```diff
@@ -84,7 +83,13 @@ protected void configure(HttpSecurity http) throws Exception {
             .accessDeniedHandler((req, resp, exc) -> resp.sendError(SC_FORBIDDEN, "You don't have authorities."))
             .and()
             .authorizeRequests()
-            .antMatchers("/management/**", "/css/**", "/img/**").hasRole(ADMIN)
+            .antMatchers(
+                "/management/**",
+                "/css/**",
+                "/img/**",
```
vorfolomey, line 89: Static files must be accessible for all

vorfolomey (changes requested):

core/src/main/java/greencity/config/SecurityConfig.java
Yura09: I didn’t change it. Those files already have been there.

vorfolomey (approved):

antonbakalets (approved):

SashkoMolodec (approved):

PR #1481, opened 2020-09-24
Graphs performance update

eyouu:


TarasDovganyuk (approved):

Yura09 (approved):

antonbakalets (changes requested):
As discussed this is not ready.

core/src/main/java/greencity/service/impl/GraphServiceImpl.java
```diff
@@ -78,17 +72,9 @@
...
-        for (User tempUser : users) {
-            if (tempUser.getDateOfRegistration().getYear() == currentYear) {
-                int month = tempUser.getDateOfRegistration().getMonth().ordinal();
-                map.put(month, map.get(month) + 1);
-            }
-        }
-
+        List<Integer> months = userRepo.findAllRegistrationMonths();
+        Map<Integer, Integer> map = initializeMapWithMonths(new TreeMap<>());
+        months.forEach(month -> map.put(month, map.get(month) + 1));
```
antonbakalets, line 77: Calculating COUNT or SUM in the database query would be even more performant.

antonbakalets (approved):

vorfolomey (approved):

PR #1482, opened 2020-09-24
List of social network images

forestik:


core/src/main/java/greencity/repository/SocialNetworkImageRepo.java
```diff
@@ -13,4 +18,26 @@
...
     Optional<SocialNetworkImage> findByHostPath(String hostPath);
+
+    /**
+     * Find {@link SocialNetworkImage} by page.
+     *
+     * @param pageable pageable configuration.
+     * @return {@link Page}
+     * @author Orest Mamchuk
+     */
+    Page<SocialNetworkImage> findAll(Pageable pageable);
```
antonbakalets, line 29: Method `findAll(Pageable)` is already available in JpaRepository that this interface extends, so no need to introduce it here.

forestik: I know this, however I noticed it in UserRepo 

image

antonbakalets (approved):

TarasDovganyuk (approved):

Yura09 (approved):

PR #1522, opened 2020-09-29
Move repository to DAO module

eyouu:


Yura09 (approved):

vorfolomey (approved):

core/src/main/java/greencity/controller/HabitController.java
```diff
@@ -5,11 +5,7 @@
 import greencity.annotations.ValidLanguage;
 import greencity.constant.HttpStatuses;
 import greencity.dto.PageableDto;
-import greencity.dto.habitstatistic.AddHabitStatisticDto;
-import greencity.dto.habitstatistic.HabitDictionaryTranslationsDto;
-import greencity.dto.habitstatistic.HabitItemsAmountStatisticDto;
-import greencity.dto.habitstatistic.HabitStatisticDto;
```
antonbakalets, line 11: `HabitStatisticDto` should stay. The constructor that depends on entity should be replaced by mapper.

antonbakalets: Most probably the same for other DTOs that are used in controllers. Only DTOs that are mapped as query result can go to dao level.

antonbakalets (changes requested):
AppConstant should definitively stay at controller (main application) level.

antonbakalets:
Here is the problem with this task. The only field from `AppConstant` used is `greencity.constant.AppConstant#DATE_FORMAT` But the conversion from String->LocalDateTime should happen at Controller level, preferably it should be Spring responsibility.

- Please try to change the greencity.dto.filter.FilterPlaceDto#time field type to LocalDateTime.

- if that is quite complicated we could hardcode the date format and create a separate ticket to fix it.

eyouu:
> Here is the problem with this task. The only field from `AppConstant` used is `greencity.constant.AppConstant#DATE_FORMAT` But the conversion from `String->LocalDateTime` should happen at Controller level, preferably it should be Spring responsibility.
> 
> * Please try to change the `greencity.dto.filter.FilterPlaceDto#time` field type to `LocalDateTime`.
> * if that is quite complicated we could hardcode the date format and create a separate ticket to fix it.

For out now just hardcoded the format, but I'll create a separate ticket and will do this task (as it was discussed)

antonbakalets (approved):

TarasDovganyuk (approved):

PR #1533, opened 2020-09-30
Sonar code smells

forestik:
…ssion


core/src/main/java/greencity/validator/TipsAndTricksDtoRequestValidator.java
```diff
@@ -25,12 +25,10 @@ public boolean isValid(TipsAndTricksDtoRequest dto, ConstraintValidatorContext c
         if (dto.getSource() != null && !dto.getSource().isEmpty()) {
             isUrlValid(dto.getSource());
         }
-        if (tagService.isValidNumOfUniqueTags(dto.getTags())) {
-            if (tagService.isAllTipsAndTricksValid(dto.getTags())) {
-                return true;
-            }
+        if ((Boolean.TRUE.equals(tagService.isValidNumOfUniqueTags(dto.getTags())))
```
antonbakalets, line 28: `Boolean.TRUE.equals(x)` is actually equal to `x`, so it can be removed.

core/src/main/java/greencity/repository/AdviceTranslationRepo.java
```diff
@@ -32,5 +32,5 @@
      * @return {@link AdviceTranslation} in Optional
      * @author Vitaliy Dzen
      */
-    Optional<AdviceTranslation> findAdviceTranslationByLanguage_CodeAndAdvice(String languageCode, String advice);
+    Optional<AdviceTranslation> findAdviceTranslationByLanguageCodeAndAdvice(String languageCode, String advice);
```
antonbakalets, line 35: Have you checked that the generated SQL query is still the same?

antonbakalets (changes requested):

core/src/main/java/greencity/validator/TipsAndTricksDtoRequestValidator.java
forestik: I understand this, but the sonar gives out code smells

antonbakalets: I believe Sonar is trying to say that `greencity.service.impl.TagsServiceImpl#isValidNumOfUniqueTags` method return type can be changed to primitive `boolean`.

forestik: I thought the same and changed to a primitive, but nothing changed.
Also in the example is :
image

forestik: Ок

core/src/main/java/greencity/validator/TipsAndTricksDtoRequestValidator.java
```diff
@@ -25,12 +25,10 @@ public boolean isValid(TipsAndTricksDtoRequest dto, ConstraintValidatorContext c
         if (dto.getSource() != null && !dto.getSource().isEmpty()) {
             isUrlValid(dto.getSource());
         }
-        if (tagService.isValidNumOfUniqueTags(dto.getTags())) {
-            if (tagService.isAllTipsAndTricksValid(dto.getTags())) {
-                return true;
-            }
+        if ((tagService.isValidNumOfUniqueTags(dto.getTags()))
+                && (!tagService.isAllTipsAndTricksValid(dto.getTags()))) {
```
antonbakalets, line 29: I'm afraid the logic has change now.

antonbakalets (changes requested):

antonbakalets (approved):

eyouu (approved):

TarasDovganyuk (approved):

PR #1564, opened 2020-10-03
Management social network images

forestik:


Yura09 (approved):

core/src/main/java/greencity/service/impl/SocialNetworkImageServiceImpl.java
```diff
@@ -106,6 +111,86 @@ public SocialNetworkImage getSocialNetworkImageByUrl(String url) {
...
+
+    /**
+     * {@inheritDoc}
+     * Method for finding {@link SocialNetworkImage} by id
+     *
+     * @param id {@link SocialNetworkImage} instance id.
+     * @return {@link SocialNetworkImage}
+     */
+    @Override
+    public SocialNetworkImage findById(Long id) {
```
antonbakalets, line 159: Could you please hide this method by making it private and removing from the interface. 

This way you will not expose `SocialNetworkImage` to the Controller layer. This is what we are trying to achieve in this epic about 3-layers application.

antonbakalets (changes requested):

antonbakalets (approved):

eyouu (approved):

PR #1570, opened 2020-10-05
Updated insert.sql

eyouu:


Yura09 (approved):

TarasDovganyuk (approved):

antonbakalets (changes requested):
Please move it into `dao/src/main/resources/db`.

Please exclude it from build, so it is not delivered inside the jar file (https://maven.apache.org/plugins/maven-resources-plugin/examples/include-exclude.html)

Feel free to use it in Repository tests.

antonbakalets:
Please consider also moving into` src/test/resources/sql`, this way there will be no need to exclude it as it will be part of test data.

eyouu:
> Please consider also moving into` src/test/resources/sql`, this way there will be no need to exclude it as it will be part of test data.

Ok. Did it this way.

antonbakalets (approved):

PR #1576, opened 2020-10-05
Delete user profile picture

forestik:


core/src/main/java/greencity/service/UserService.java
```diff
@@ -298,6 +299,14 @@
     User updateUserProfilePicture(MultipartFile image, String email,
                                   UserProfilePictureDto userProfilePictureDto);
 
+    /**
+     * Delete user profile picture {@link User}.
+     *
+     * @param email {@link String} - email of user that need to update.
+     * @return {@link User}.
+     */
+    User deleteUserProfilePicture(String email);
```
antonbakalets, line 308: Can you please do not return `User`, in order not to expose `User` to controllers layer.

antonbakalets (changes requested):

core/src/main/java/greencity/service/UserService.java
forestik: Done

antonbakalets (approved):

Yura09 (approved):

eyouu (approved):

PR #1583, opened 2020-10-06
Fix some sonar issues

vorfolomey:


core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -11,6 +11,7 @@
 import java.util.stream.Collectors;
 import javax.validation.ConstraintViolation;
 import javax.validation.ConstraintViolationException;
+import javax.validation.constraints.NotNull;
```
antonbakalets, line 14: This annotation will nor work here.

antonbakalets (approved):

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
vorfolomey: Changed to Lombok NotNull

antonbakalets (approved):

forestik (approved):

PR #1587, opened 2020-10-06
Statistics update

eyouu:


Yura09 (approved):

dao/src/main/java/greencity/repository/UserRepo.java
```diff
@@ -220,9 +223,19 @@
...
+    @Query(nativeQuery = true)
+    List<RegistrationStatisticsDtoResponse> findAllRegistrationMonths();
+
+    /**
+     * Converts result of findAllRegistrationMonths() method to {@link Map}.
+     *
+     * @return {@link Map}
+     */
+    default Map<Integer, Long> findAllRegistrationMonthsMap() {
+        return findAllRegistrationMonths().stream().collect(Collectors.groupingBy(
```
antonbakalets, line 237: Why do you need grouping?
I think, you already have 12 unique elements in the query result. You just need to convert it into a map, for example

`return findAllRegistrationMonths().stream().collect(Collectors.toMap(RegistrationStatisticsDtoResponse::getMonth, RegistrationStatisticsDtoResponse::getCount));`

dao/src/main/java/greencity/repository/UserRepo.java
```diff
@@ -220,9 +223,19 @@
     /**
      * Find and return all registration months.
      *
-     * @return {@link List} of months
+     * @return {@link List} of {@link RegistrationStatisticsDtoResponse}
```
antonbakalets, line 226: This JavaDoc would be much useful if it can explain that the query behind this method is the one in 
```
@NamedNativeQuery(name = "User.findAllRegistrationMonths"
```
and how this method identifies it.

forestik (approved):

antonbakalets (changes requested):

dao/src/main/java/greencity/repository/UserRepo.java
eyouu: Yeah, did it by mistake.

dao/src/main/java/greencity/repository/UserRepo.java
eyouu: Done. Does it look good now?

dao/src/main/java/greencity/dto/user/RegistrationStatisticsDtoResponse.java
```diff
@@ -0,0 +1,14 @@
+package greencity.dto.user;
+
+import lombok.*;
+
+@Getter
+@Setter
+@NoArgsConstructor
```
antonbakalets, line 7: **Comment for everyone.**

One more thing that I noticed, almost in each pull request you are putting almost all Lombok annotations. But here, for example, I would remove `@NoArgsConstructor` and `@Builder` and even `@Setter`

dao/src/main/java/greencity/repository/UserRepo.java
antonbakalets: Much better now.

antonbakalets (approved):

PR #1601, opened 2020-10-07
Filtering tips and tricks

Yura09:


core/src/main/java/greencity/webcontroller/ManagementTipsAndTricksController.java
```diff
@@ -149,4 +161,22 @@ public GenericResponseDto save(@Valid
...
+     * @param model                ModelAndView that will be configured and returned to user.
+     * @param tipsAndTricksViewDto used for receive parameters for filters from UI.
+     */
+    @PostMapping(value = "", consumes = MediaType.APPLICATION_FORM_URLENCODED_VALUE)
+    public String filterData(Model model,
+                             @PageableDefault(value = 20) @ApiIgnore Pageable pageable,
+                             TipsAndTricksViewDto tipsAndTricksViewDto) {
+        PageableDto<TipsAndTricksDtoResponse> pageableDto =
+            tipsAndTricksService.getFilteredDataForManagementByPage(pageable,
+                tipsAndTricksService.getSpecification(tipsAndTricksViewDto));
```
antonbakalets, line 177: Conversion from `tipsAndTricksViewDto` to `Specification` should happen in `TipsAndTricksServiceImpl`.

`Specification` (and its implementations) is something database layer specific, so only Service Implementations can access it.

In one of my project I practiced to put method for `Specification` creation as default method of a `Repo` interface.

antonbakalets (changes requested):

antonbakalets (approved):

TarasDovganyuk (approved):

forestik (approved):

PR #1610, opened 2020-10-08
issue1609_spesification

TarasDovganyuk:
Shared methods moved to Interface


core/src/main/java/greencity/filters/MySpecification.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.Predicate;
+import javax.persistence.criteria.Root;
+import org.springframework.context.annotation.Scope;
+import org.springframework.data.jpa.domain.Specification;
+import org.springframework.stereotype.Component;
+
+@Component
+@Scope("prototype")
+public interface MySpecification<T> extends Specification<T> {
```
antonbakalets, line 16: Why interface? Why is this a @Component?

What I was explaining is, for example:
- you have `RatingStatisticsRepo extends JpaSpecificationExecutor<RatingStatistics>` -> allows you to call `findAll(Specification)`
- `RatingStatisticsRepo` is an interface -> you can't add logic
- but starting from Java 8 you can add `default` methods -> put the method that converts DTO to `Specification` in the interface/repo that uses this Specification

antonbakalets (changes requested):

antonbakalets (approved):

forestik (approved):

eyouu (approved):

Issue #1580, opened 2020-10-06
Add and edit Tips&tricks in 3 languages in web Admin


olenapetryshak:
Description
As an admin, I want to add and edit the tips and tricks(title and text) in 3 languages.

PR #1620, opened 2020-10-11
Add and edit tips&tricks in 3 languages 1580

olenapetryshak:
https://github.com/ita-social-projects/GreenCity/issues/1580


core/src/main/java/greencity/dto/tipsandtricks/TextTranslationDTO.java
```diff
@@ -0,0 +1,22 @@
...
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class TextTranslationDTO {
+    @NotEmpty
+    @Size(min = 20, max = 63206)
```
antonbakalets, line 18: I would rather put 4000. Where does 63206 number come from?

dao/src/main/java/greencity/entity/TextTranslation.java
```diff
@@ -0,0 +1,26 @@
...
+
+import javax.persistence.*;
+
+@Entity
+@Table(name = "text_translations")
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class TextTranslation {
```
antonbakalets, line 13: `TextTranslation` and `TitleTranslation` could extend some abstract class `Translation` as they have common fields. It would be interesting to map them to they tables. Please consider as separate ticket.

antonbakalets (approved):

forestik (approved):

SashkoMolodec (approved):

PR #1651, opened 2020-10-16
update method for EcoNewsController

Yura09:


core/src/main/java/greencity/dto/econews/UpdateEcoNewsDto.java
```diff
@@ -0,0 +1,40 @@
...
+@Builder
+public class UpdateEcoNewsDto {
+    @NotNull
+    @Min(1)
+    private Long id;
+
+    @Size(min = 1, max = 170)
+    private String title;
+
+    @Size(min = 20, max = 63206)
```
antonbakalets, line 31: This looks too much.

Yura09: But entity also has the same constraints

forestik (approved):

eyouu (approved):

core/src/main/java/greencity/dto/econews/UpdateEcoNewsDto.java
antonbakalets: By the way, is constraint on entity is still working or can be removed? 

You will need to disable this one to check.

Yura09: done.

TarasDovganyuk (approved):

Issue #1709, opened 2020-10-22
Refactor update method in HabitServiceImpl.


SashkoMolodec:
Better move this from lambda. Make getHabitTranslations rerutn Map<String, HabitTranslationManagementDto>, where key will be lamguage. Than find needed HabitTranslationManagementDto using get in Map.

_Originally posted by @vterlyha in https://github.com/ita-social-projects/GreenCity/pull/1700#discussion_r510102289_

Issue #1708, opened 2020-10-22
refactor methods in HabitStatusServiceImpl.


SashkoMolodec:
This method is really huge and needs refatoring. First, please, move if else statements to private methods and than we will se how can we esle refator it.

_Originally posted by @vterlyha in https://github.com/ita-social-projects/GreenCity/pull/1700#discussion_r510103685_

SashkoMolodec:
I won't refactor other methods in HabitStatusServiceImpl because maybe they'll probably become unused in feature or they will be rewrited fully.

PR #1729, opened 2020-10-27
Refactor methods in habits.

SashkoMolodec:
https://github.com/ita-social-projects/GreenCity/issues/1709
https://github.com/ita-social-projects/GreenCity/issues/1708


Markiyan000 (approved):

Yura09 (approved):

service/src/main/java/greencity/service/HabitServiceImpl.java
```diff
@@ -163,23 +161,40 @@ public void update(HabitManagementDto habitManagementDto, MultipartFile image) {
...
         habitRepo.save(habit);
     }
 
+    /**
+     * Method updates {@link HabitTranslation} with {@link HabitTranslationManagementDto} fields.
+     *
+     * @param htDto {@link HabitTranslationManagementDto} instance.
+     * @param ht    {@link HabitTranslation} instance.
+     */
+    private void updateHabitTranslation(HabitTranslationManagementDto htDto, HabitTranslation ht) {
```
vterlyha, line 179: bad method naming, update is used when smth is updatedin DB, for example. Please, rename it to smth like enhanceWith...

core/src/main/java/greencity/mapping/HabitStatusDtoMapper.java
```diff
@@ -27,11 +28,11 @@ protected HabitStatusDto convert(HabitStatus habitStatus) {
             .lastEnrollmentDate(habitStatus.getLastEnrollmentDate())
             .workingDays(habitStatus.getWorkingDays())
             .habitAssignId(habitStatus.getHabitAssign().getId())
-            .habitStatusCalendarDtos(habitStatus.getHabitStatusCalendars().stream()
+            .habitStatusCalendarDtos(new ArrayList<>(habitStatus.getHabitStatusCalendars().stream()
```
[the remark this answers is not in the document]
SashkoMolodec: In that case I'm getting exception in tests. And I don't now if it's even possible to fix it. It's hard for me to remove this line with adding, because I need to return this dto as a response. So I came with ArrayList alternative to List.
Знімок екрана 2020-10-28 162029

service/src/main/java/greencity/service/HabitServiceImpl.java
SashkoMolodec: enhanceTranslationWithDto - would be great?

vterlyha (approved):

Issue #1612, opened 2020-10-08
Generate Hibernate metamodel


antonbakalets:
**As a** developer **I want to** have Hibernate Metamodel generated **so that** I can build Specifications

**Description**
Add JPAMetaModelEntityProcessor to DAO module.
See https://docs.jboss.org/hibernate/jpamodelgen/1.0/reference/en-US/html_single/
You might also need https://stackoverflow.com/a/31805414

### Acceptance Criteria
After running `mvn compile` on DAO module additional classes are generated in `dao/target/generated-sources/annotations/greencity/entity`

PR #1821, opened 2020-11-11
Hibernate metamodel created

IvanBegar:
https://github.com/ita-social-projects/GreenCity/issues/1612


forestik (approved):

KhonkoDmytro (approved):

SashkoMolodec (approved):

service/src/main/java/greencity/service/HabitFactServiceImpl.java
```diff
@@ -121,18 +124,20 @@ public HabitFactVO save(HabitFactPostDto fact) {
...
+        habitFact.setHabit(habit);
+        habitFact.getTranslations()
+            .forEach(habitFactTranslation -> {
+                habitFactTranslation.setContent(factDto.getTranslations().stream()
+                    .filter(newTranslation -> newTranslation.getLanguage().getCode()
+                        .equals(habitFactTranslation.getLanguage().getCode()))
+                    .findFirst().get()
+                    .getContent());
+                habitFactTranslation.setFactOfDayStatus(factDto.getTranslations().get(0).getFactOfDayStatus());
+            });
```
vterlyha, line 139: Move it to private method with proper name

dao/src/main/java/greencity/entity/HabitFact.java
```diff
@@ -17,9 +17,9 @@
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
 
-    @OneToMany(cascade = {CascadeType.ALL}, mappedBy = "habitFact", fetch = FetchType.LAZY)
+    @OneToMany(cascade = {CascadeType.ALL}, mappedBy = "habitFact", fetch = FetchType.EAGER)
     private List<HabitFactTranslation> translations;
 
-    @ManyToOne
+    @ManyToOne(fetch = FetchType.EAGER)
```
vterlyha, line 23: Use EAGER only when every request in your application needs this filed. In other case use LAZY and if you need this field  when connection to DB is closed use custom query with inner join

service/src/main/java/greencity/service/HabitFactServiceImpl.java
IvanBegar: done

dao/src/main/java/greencity/entity/HabitFact.java
IvanBegar: fixed

PR #1972, opened 2020-12-03
tags-localization-1944

Markiyan000:


dao/src/main/java/greencity/entity/localization/TagTranslation.java
```diff
@@ -0,0 +1,31 @@
+package greencity.entity.localization;
+
+import greencity.entity.Language;
+import greencity.entity.Tag;
+import lombok.*;
+
+import javax.persistence.*;
+
+@Data
```
vterlyha, line 9: Do you really need data here ?

dao/src/main/java/greencity/repository/EcoNewsRepo.java
```diff
@@ -54,6 +55,16 @@
...
     Page<EcoNews> findAllByOrderByCreationDateDesc(Pageable page);
 
+    /**
+     * Method that finds {@link EcoNews} by id.
+     * 
+     * @param id {@link Long}.
+     * @return {@link Optional} of {@link EcoNews}
+     *
+     */
+    @Query("select e from EcoNews e join fetch e.tags where e.id = :id")
```
vterlyha, line 65: It is a good approach to have similar code style to all queries. I would prefer UPPER CASE for key words as SELECT, WHERE etc.

dao/src/main/java/greencity/repository/HabitTranslationRepo.java
```diff
@@ -77,8 +77,8 @@
      */
     @Query(nativeQuery = true, value = "select distinct ht.* from habit_translation as ht "
         + "inner join habits_tags as htg on ht.habit_id = htg.habit_id "
-        + "inner join tags as t on t.id = htg.tag_id "
+        + "inner join tag_translations as t on t.tag_id = htg.tag_id "
```
vterlyha, line 80: Refactor in all queries, please

forestik (approved):

SashkoMolodec (approved):

KhonkoDmytro (approved):

PR #2688, opened 2021-04-28
Fixed bug with user's shopping list on 'My habits' page.

BOBAH123:
Now there we will get a list of users Active shopping list items from his habits which are in progress.
image

### DIFF
--- a/service/src/main/java/greencity/service/CustomShoppingListItemServiceImpl.java
+++ b/service/src/main/java/greencity/service/CustomShoppingListItemServiceImpl.java
@@ -2,8 +2,10 @@ package greencity.service;
 
 import greencity.client.RestClient;
 import greencity.constant.ErrorMessage;
-import static greencity.constant.ErrorMessage.CUSTOM_SHOPPING_LIST_ITEM_NOT_FOUND_BY_ID;
-import greencity.dto.shoppinglistitem.*;
+import greencity.dto.shoppinglistitem.BulkSaveCustomShoppingListItemDto;
+import greencity.dto.shoppinglistitem.CustomShoppingListItemResponseDto;
+import greencity.dto.shoppinglistitem.CustomShoppingListItemSaveRequestDto;
+import greencity.dto.shoppinglistitem.ShoppingListItemDto;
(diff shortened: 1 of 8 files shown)


hutei (approved):

ZaharVeremchuk (approved):

V-la-d (approved):

service/src/main/java/greencity/service/CustomShoppingListItemServiceImpl.java
```diff
@@ -203,6 +201,15 @@ public CustomShoppingListItemResponseDto updateItemStatus(Long userId, Long item
             }.getType());
     }
 
+    @Override
+    public List<ShoppingListItemDto> findByActiveByUserIdAndLanguageCode(Long userId, String code) {
+        List<ShoppingListItemDto> shoppingListItems = new ArrayList<>();
+        customShoppingListItemRepo.findByActiveByUserIdAndLanguageCode(userId, code)
+            .forEach(x -> shoppingListItems.add(modelMapper.map(
```
stas-mureiko, line 208: Would it better to use stream().map() instead of .foreach()? This way you will not need to use the new list

service/src/test/java/greencity/service/CustomShoppingListItemServiceImplTest.java
```diff
@@ -251,6 +246,13 @@ void updateItemStatus() {
 
     }
 
+    @Test
+    void findByActiveByUserIdAndLanguageCodeTest() {
+        when(customShoppingListItemService.findByActiveByUserIdAndLanguageCode(1L, "ua"))
```
stas-mureiko, line 251: please, be mindful what you stub and how this stub use

stas-mureiko (commented):
Left some comments. Other code looks good

service/src/main/java/greencity/service/CustomShoppingListItemServiceImpl.java
BOBAH123: Changed like this 
image

PR #3878, opened 2022-02-15
refactoring ApiResponse

Zingerd:
refactoring ApiResponse. Deleted not relevant  api response


Max877866 (approved):

Andrii-Yezenitskyi (approved):

dessdee (approved):

thesvider (commented):
Good. if API doesn't return those specific codes -we should not include them into the documentation

Issue #4542, opened 2022-08-15
[API] [Eco news] Redundant status codes 400 and 403 in the documentation of Swagger when getting eco news by id


TaniaIvashkiv:
**Environment:** all.
**Reproducible:** always.
**Build found:** 15.08.22

**Preconditions**
1. Postman is opened

**Steps to reproduce**
1. Make method GET greencity.testgreencity.ga/econews/{id}

**Actual result**
1. The status codes 400 and 403 are redundant in the documentation of Swagger

**Expected result**
1. Can get only 200 and 404 status codes performing request GET greencity.testgreencity.ga/econews/{id}

**User story and test case links**
E.g.: "User story #498
[Test case](https://jira.softserve.academy/browse/100)"

Issue #4547, opened 2022-08-18
[API] [Eco news] Redundant status codes 400 and 404 in the documentation of Swagger when user wants to get 3 last eco news


mikhaylovamarina:
**Environment:** all
**Reproducible:** always
**Build found:** 18.08.2022

**Preconditions**
Postman is opened
Swagger is opened https://greencity.testgreencity.ga/swagger-ui.html#/ ([eco-news-controller](https://greencity.testgreencity.ga/swagger-ui.html#/eco-news-controller))

**Steps to reproduce**
1. Make method GET greencity.testgreencity.ga/econews/newest

**Actual result**
Can get only 200 status code performing request GET greencity.testgreencity.ga/econews/newest

**Expected result**
The status codes 400 and 404 are redundant in the documentation of Swagger

**Labels to be added**
"Bug", Priority ("law: "), Severity ("trivial:"), Type ("Functional"), "API" (for back-end bugs).

PR #4548, opened 2022-08-18
removed redundant status codes (#4542)

OlegVat:
Oleh Vatuliak
issue #4542

## Summary of issue

Removed redundant status code

## Summary of change

Removed redundant 403 code

### DIFF
--- a/core/src/main/java/greencity/controller/EcoNewsController.java
+++ b/core/src/main/java/greencity/controller/EcoNewsController.java
@@ -175,8 +175,6 @@ public class EcoNewsController {
     @ApiOperation(value = "Get eco news by id.")
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = HttpStatuses.OK),
-        @ApiResponse(code = 400, message = HttpStatuses.BAD_REQUEST),
-        @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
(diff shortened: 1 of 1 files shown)


danylohlynsky (approved):

innayashna (approved):

core/src/main/java/greencity/controller/EcoNewsController.java
```diff
@@ -175,8 +175,6 @@ public ResponseEntity<EcoNewsGenericDto> update(
     @ApiOperation(value = "Get eco news by id.")
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = HttpStatuses.OK),
-        @ApiResponse(code = 400, message = HttpStatuses.BAD_REQUEST),
```
danylohlynsky, line 178: It should be possible to get a 400 response when someone tries to enter char as id

OlegVat: https://github.com/ita-social-projects/GreenCity/issues/4547

This issue says that 404 is also redundant but used in the if statement in the implementation (so I left that code).

So does this ticket make any sense?

OlegVat: Fixed

AndriyUrban (approved):

NataliaHlazova (approved):

PR #4720, opened 2022-11-07
Event is active field, Delete the event

kmpol:
Before the (partial - no tests) implementation:

Delete event button: no behavior

After the (partial - no tests) implementation:

Delete button deletes in two way:
-- First: If there are event attenders - soft delete (remove from view, but still available in the db)
-- Second: If there are no event attenders - hard delete (remove from db)

Before the deletion - modal asks if you are sure about the deletion.

### DIFF
--- a/core/src/main/java/greencity/webcontroller/ManagementEventsController.java
+++ b/core/src/main/java/greencity/webcontroller/ManagementEventsController.java
@@ -59,8 +59,8 @@ public class ManagementEventsController {
     }
 
     /**
-     * Method that perform soft or hard delete of the event.
-     * If no attenders are signed to an event then hard delete, if they are, then soft delete.
+     * Method that perform soft or hard delete of the event. If no attenders are
+     * signed to an event then hard delete, if they are, then soft delete.
      *
      * @return View template path {@link String}.
      */
--- a/core/src/main/java/greencity/webcontroller/ManagementEventsController.java
+++ b/core/src/main/java/greencity/webcontroller/ManagementEventsController.java
@@ -30,7 +30,7 @@ public class ManagementEventsController {
      */
     @GetMapping
     public String getAllEvents(@RequestParam(required = false, name = "query") String query, Model model,
-        @ApiIgnore Pageable pageable, EventViewDto eventViewDto, @RequestParam(required = false, name = "active", defaultValue = "false") boolean active) {
+        @ApiIgnore Pageable pageable, EventViewDto eventViewDto) {
         PageableAdvancedDto<EventDto> allEvents;
         if (!eventViewDto.isEmpty()) {
             allEvents = eventService.getAll(pageable, null);
@@ -58,8 +58,14 @@ public class ManagementEventsController {
         return "core/management_events";
     }
 
+    /**
+     * Method that perform soft or hard delete of the event.
+     * If no attenders are signed to an event then hard delete, if they are, then soft delete.
+     *
+     * @return View template path {@link String}.
+     */
     @PostMapping("/delete")
-    public String deleteEvent(@RequestParam Long id){
+    public String deleteEvent(@RequestParam Long id) {
         eventService.disableEvent(id);
         return "redirect:/management/events";
     }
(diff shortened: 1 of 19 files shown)


core/src/main/java/greencity/webcontroller/ManagementEventsController.java
```diff
@@ -30,7 +30,7 @@ public class ManagementEventsController {
      */
     @GetMapping
     public String getAllEvents(@RequestParam(required = false, name = "query") String query, Model model,
-        @ApiIgnore Pageable pageable, EventViewDto eventViewDto) {
+        @ApiIgnore Pageable pageable, EventViewDto eventViewDto, @RequestParam(required = false, name = "active", defaultValue = "false") boolean active) {
```
marcinkunert: If you use defaultValue then `required = false` is not needed

core/src/main/resources/static/css/sidebar.css
```diff
@@ -199,7 +199,18 @@ block:hover img{
     color: white;
 }
 
+button.close {
+    padding: 2rem 1rem !IMPORTANT;
```
marcinkunert, line 203: I've never seen `!important` written in caps 😅

core/src/main/resources/templates/core/management_events.html
```diff
@@ -355,29 +384,29 @@ <h4 class="modal-title">[[#{greenCity.events.page.edit.events}]]</h4>
...
+<!--                            <input type="button" class="btn btn-default btn-tertiary" data-dismiss="modal"-->
+<!--                                   th:value="#{greenCity.pages.popup.cancel.}">-->
+<!--                            <a href="" id="deleteOneSubmit" class="btn btn-remove">-->
+<!--                                <div>[[#{greenCity.pages.delete}]]</div>-->
+<!--                            </a>-->
+<!--                        </div>-->
+<!--                    </form>-->
+<!--                </div>-->
+<!--            </div>-->
+<!--        </div>-->
```
marcinkunert: Did you comment that out on purpose? Either delete it completely or bring this back

service/src/main/java/greencity/service/EventServiceImpl.java
```diff
@@ -76,6 +77,7 @@ public EventDto save(AddEventDtoRequest addEventDtoRequest, String email,
     }
 
     @Override
+    @Transactional
```
marcinkunert: If this is added on purpose, what's the benefit?

service/src/main/java/greencity/service/EventServiceImpl.java
```diff
@@ -136,6 +138,7 @@ private void setSubscribes(Page<Event> events, PageableAdvancedDto<EventDto> eve
 
     private PageableAdvancedDto<EventDto> buildPageableAdvancedDto(Page<Event> eventsPage) {
         List<EventDto> eventDtos = eventsPage.stream()
+                .filter(Event::getIsActive)
```
marcinkunert: It is better to do filtering on the database side.

service/src/main/java/greencity/service/EventServiceImpl.java
```diff
@@ -254,6 +257,21 @@ public EventVO findById(Long eventId) {
...
+        //TODO: remove duplicated code
+        Set<EventAttenderDto> allEventAttenders = getAllEventAttenders(eventId);
+        UpdateEventDto eventUpdates = new UpdateEventDto();
+        eventUpdates.setId(eventId);
+        eventUpdates.setIsActive(false);
+        if(allEventAttenders.isEmpty()){
+            delete(eventId, (String) SecurityContextHolder.getContext().getAuthentication().getPrincipal());
+        } else {
+            update(eventUpdates, (String) SecurityContextHolder.getContext().getAuthentication().getPrincipal(), null);
+        }
```
marcinkunert: I'm not a fan of this approach. Consider moving the principal part into delete and update metods

marcinkunert (changes requested):

service/src/main/java/greencity/service/EventServiceImpl.java
kmpol: Yes, but there might be in the future some functionality, which could be "filter", that shows me only active and inactive events.
For now we are getting whole db and filter inactive events out. In the future we still want whole db downloaded and just filter active out in order to show inactive ones. So we could active them back.

PawelMiskowiec (approved):

PR #6893, opened 2024-01-16
implemented filtering for favorite events

0lia:
# GreenCity PR

## Added
- filtering for favorite events


holotsvan:
filtering for favorite events are implemented, so there no need in this old pull request, so close it.

PR #6900, opened 2024-01-19
show the list of favorite ecoNews to user

0lia:
# GreenCity PR

## Added
- get all favorite ecoNews for the current user


holotsvan:
Finding all  favorite ecoNews is implemented in findAll method, so no need in this functionality, so that close this pull request

PR #7474, opened 2024-09-19
Fix events

IliaWithHat:
## Summary of change

- Right working filtering by statuses, tags and cities.


core/src/main/java/greencity/controller/EventController.java
```diff
@@ -168,13 +171,22 @@ public ResponseEntity<EventDto> getEvent(
...
         @Parameter(hidden = true) Pageable pageable,
         @RequestParam(required = false, name = "user-id") Long userId,
         FilterEventDto filterEventDto) {
+        if (filterEventDto != null && filterEventDto.getStatuses() != null) {
+            validateStatusesRequireUserId(filterEventDto.getStatuses(), userId);
+        }
         return ResponseEntity.ok().body(eventService.getEvents(pageable, filterEventDto, userId));
     }
 
+    private void validateStatusesRequireUserId(List<EventStatus> statuses, Long userId) {
```
nazariusPr, line 184: Maybe it is better to transfer this logic into service layer

dao/src/main/java/greencity/repository/impl/EventSearchRepoImpl.java
```diff
@@ -137,7 +137,7 @@ private void addCitiesPredicate(List<String> cities, Root<Event> eventRoot, List
         if (cities != null && !cities.isEmpty()) {
             Join<EventDateLocation, Address> addressJoin = eventRoot
                 .join(Event_.dates, JoinType.LEFT).join(EventDateLocation_.address);
-            predicates.add(addressJoin.get(Address_.CITY_EN).in(cities));
+            predicates.add(criteriaBuilder.upper(addressJoin.get(Address_.CITY_EN)).in(cities));
```
nazariusPr: it is also a good idea to make cities be in upper case too

dao/src/main/java/greencity/repository/impl/EventSearchRepoImpl.java
```diff
@@ -170,7 +170,7 @@ private void addStatusesPredicate(List<EventStatus> eventStatuses, Long userId,
     private void addTagsPredicate(List<String> tags, Root<Event> eventRoot, List<Predicate> predicates) {
         if (tags != null && !tags.isEmpty()) {
             ListJoin<Tag, TagTranslation> tagsJoin = eventRoot.join(Event_.tags).join(Tag_.tagTranslations);
-            predicates.add(tagsJoin.get(TagTranslation_.NAME).in(tags));
+            predicates.add(criteriaBuilder.upper(tagsJoin.get(TagTranslation_.NAME)).in(tags));
```
nazariusPr: it is also a good idea to make tags be in upper case too

core/src/main/java/greencity/controller/EventController.java
```diff
@@ -168,13 +171,22 @@ public ResponseEntity<EventDto> getEvent(
     })
     @ApiPageableWithoutSort
     @GetMapping
-    public ResponseEntity<PageableAdvancedDto<EventDto>> getEvent(
+    public ResponseEntity<PageableAdvancedDto<EventDto>> getEvents(
         @Parameter(hidden = true) Pageable pageable,
         @RequestParam(required = false, name = "user-id") Long userId,
         FilterEventDto filterEventDto) {
```
nazariusPr, line 177: Please also look at FilterEventDto class, because it is expected that there will be List of EventTime (not just EventTime). Due to this problem event filter doesnt work on client side, when user selects several event times

IliaWithHat: No. EventTime can be only FUTURE or PAST. I discuss this with Pavlo Poliak.

core/src/main/java/greencity/controller/EventController.java
IliaWithHat: No. This simple validation, that do not related to business logic, should be on controller layer. For example see DTO validation using annotation from jakarta.validation.constraints package.

dao/src/main/java/greencity/repository/impl/EventSearchRepoImpl.java
IliaWithHat: Ok.

dao/src/main/java/greencity/repository/impl/EventSearchRepoImpl.java
IliaWithHat: Ok.

core/src/main/java/greencity/controller/EventController.java
IliaWithHat: If I could write this with annotations I would, but this validation requires two objects: statuses list and userId.

K4sik (approved):

KizerovDmitriy (approved):

Maryna-511750 (approved):

PR #7690, opened 2024-10-27
7678 favorite eco news

RockTF:
**_Favorites Filter Enabled:_**

When the Favorites filter is enabled, the user should receive only their favorite EcoNews entries.
If the user performs a search query (e.g., by title), the results will include only those favorite news entries whose titles match the search query.

`Important`: Even if non-favorite EcoNews entries have the same title as favorites, only favorites should be shown in the results.

**_Favorites Filter Disabled (General Search):_**

If the Favorites filter is not enabled, the search query will be performed across all available EcoNews entries, returning both favorite and non-favorite news based on the search criteria (e.g., matching title).


K4sik:
You have 3 new issues in your code.
image

ToriForH (approved):

IliaWithHat (approved):

K4sik (approved):

Issue #7662, opened 2024-10-22
[Comment Notifications] Notifications about comments do not include link for user to be redirected to the page with comment


Roksana77:
**Environment:**Windows 10, Chrome 127.0.6533.89, 64-bit.
**Reproducible:** always
**Build found:** 22.10.24

**Preconditions**
The page https://www.greencity.cx.ua/#/events is opened.
The user is logged in.

**Steps to reproduce**
1. User 1 writes comment to User's 2 event.
2. User 2 checks notification on web-site.
3. User 2 pay attention if he or she has the button 'Go to comment' to be redirected to the page with this new comment he or she has just received.

**Actual result**
The user receives notifications about comment or tag in comment without ability to be redirected to the page with this new comment. 

![image](https://github.com/user-attachments/assets/d8f57dc0-31f7-460b-a3fd-bd098f89546b)

Messages of notifications about likes for comments show the information about the author twice in case of having two likes from the author.    

The title of event is displayed as 'null'.

![image](https://github.com/user-attachments/assets/57d471b1-f97a-423b-aeb4-a15d40a53492)

**Expected result**
The user receives notifications about comment with ability to be redirected to the page with this new comment. 

(https://github.com/ita-social-projects/GreenCity/issues/4665).

**User story and test case links**
#4665

**Labels to be added**

Roksana77:
It still doesn't work for likes.

MilaKomisarenko:
The user isn't redirected to the event when the user's event receives a like and the user clicks on the event's name.
![image](https://github.com/user-attachments/assets/386f5d42-cda7-41bf-a386-7537fb8f041d)

PR #7744, opened 2024-11-06
Fix null notification name for comments and improve notification

fedykvitalik2004:
#7662
## Summary of changes
-Enhanced notification text
-Added support for messaging tailored to varying user group sizes
-Fixed null title


Maryna-511750 (approved):

holotsvan (approved):

service/src/main/java/greencity/service/EcoNewsServiceImpl.java
IliaWithHat, line 1: You can use overloaded methods instead of passing null parameters.

KizerovDmitriy (approved):

PR #7944, opened 2024-12-19
Bugfix/6957/fix update response event method

Cr1stal423:
GreenCity PR

Issue Link:
[6957](https://github.com/orgs/ita-social-projects/projects/43/views/1?filterQuery=label%3AAPI&pane=issue&itemId=69330933&issue=ita-social-projects%7CGreenCity%7C6957)

Changes:

    Added logic to validate that startDate and finishDate are not empty when updating an event.
    Added new error messages.
    Fixed the API response to properly handle errors.

Current Behavior:

    The response now indicates when startDate or finishDate are not provided.
    
    
image

### DIFF
--- a/core/src/main/java/greencity/validator/EventDtoRequestValidator.java
+++ b/core/src/main/java/greencity/validator/EventDtoRequestValidator.java
@@ -34,7 +34,7 @@ public class EventDtoRequestValidator implements ConstraintValidator<ValidEventD
             }
 
             for (var eventDateLocationDto : eventDateLocationDtos) {
-                if (!validateEventDateLocation(eventDateLocationDto, context)) {
+                if (!isValidEventDateLocation(eventDateLocationDto, context)) {
                     return false;
                 }
 
@@ -63,7 +63,7 @@ public class EventDtoRequestValidator implements ConstraintValidator<ValidEventD
             List<EventDateLocationDto> eventDateLocationDtos = updateEventDto.getDatesLocations();
 
             for (var eventDateLocationDto : eventDateLocationDtos) {
-                if (!validateEventDateLocation(eventDateLocationDto, context)) {
+                if (!isValidEventDateLocation(eventDateLocationDto, context)) {
                     return false;
                 }
             }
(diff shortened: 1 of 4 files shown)


core/src/main/java/greencity/validator/EventDtoRequestValidator.java
```diff
@@ -6,49 +6,89 @@
...
+
+            return true;
         }
 
-        if (value.getTags().size() > ValidationConstants.MAX_AMOUNT_OF_TAGS) {
-            throw new EventDtoValidationException(ErrorMessage.WRONG_COUNT_OF_TAGS_EXCEPTION);
+        return false;
+    }
+
+    private boolean validateEventDateLocation(EventDateLocationDto eventDateLocationDto,
```
LazarenkoDmytro: It would be more understandable if you began the boolean-returning method’s name with “isValid” instead of “validate”

LazarenkoDmytro (commented):
Left a comment

core/src/main/java/greencity/validator/EventDtoRequestValidator.java
Cr1stal423: changed the method name

LazarenkoDmytro (approved):

core/src/main/java/greencity/validator/EventDtoRequestValidator.java
```diff
@@ -6,49 +6,89 @@
...
-                throw new EventDtoValidationException(ErrorMessage.EVENT_START_DATE_AFTER_FINISH_DATE_OR_IN_PAST);
+            for (var eventDateLocationDto : eventDateLocationDtos) {
+                if (!isValidEventDateLocation(eventDateLocationDto, context)) {
+                    return false;
+                }
+
+                if (eventDateLocationDto.getStartDate().isBefore(ZonedDateTime.now())
+                    || eventDateLocationDto.getStartDate().isAfter(eventDateLocationDto.getFinishDate())) {
+                    throw new EventDtoValidationException(ErrorMessage.EVENT_START_DATE_AFTER_FINISH_DATE_OR_IN_PAST);
+                }
```
VasylyshynDmytro, line 44: this check should be moved to a separate method (not critical, not necessarily)

Cr1stal423: It was like that before, not my changes. Also, since the isValid method has a loop that already goes through the EventDateLocationDto list, in this case I do not see the need to change the method that goes through the list again

KizerovDmitriy (approved):

holotsvan (approved):

PR #7989, opened 2025-01-05
Add different profile privacy states 

Cr1stal423:
😎 GreenCity PR 😎

Issue Link:
[7970](https://github.com/orgs/ita-social-projects/projects/43/views/1?pane=issue&itemId=92306093&issue=ita-social-projects%7CGreenCity%7C7970)
Changes:

Add different profile privacy states like: PRIVATE(visible only for user), FRIENDS_ONLY(visible for user and user and their friends), PRIVATE(visible for all users)
Create new Liquibase changelog to change column type in DB and update previous values

Now Response looks like this:

image

### DIFF
--- a/dao/src/main/java/greencity/enums/EcoPlacePrivacyPolicy.java
+++ b/dao/src/main/java/greencity/enums/EcoPlacePrivacyPolicy.java
@@ -1,0 +1,7 @@
+package greencity.enums;
+
+public enum EcoPlacePrivacyPolicy {
+    PRIVATE,
+    FRIENDS_ONLY,
+    PUBLIC
+}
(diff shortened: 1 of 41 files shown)


softservedata:
Ok

softservedata (approved):

Cr1stal423:
@coderabbitai review

dao/src/main/java/greencity/enums/EcoPlacePrivacyPolicy.java
```diff
@@ -0,0 +1,7 @@
+package greencity.enums;
+
+public enum EcoPlacePrivacyPolicy {
```
KizerovDmitriy: if we have the same values in three different fields, maybe refactor this case, and left one enum?

KizerovDmitriy (changes requested):

dao/src/main/java/greencity/enums/EcoPlacePrivacyPolicy.java
Cr1stal423: I made 3 enums so that in the future, if there are different statuses for different fields, I can have them. But since they are the same now, I can probably make one

VasylyshynDmytro (approved):

KizerovDmitriy (approved):

PR #8009, opened 2025-01-08
Add different profile privacy states prod env

Cr1stal423:
😎 GreenCity PR 😎

Issue Link:
[7970](https://github.com/orgs/ita-social-projects/projects/43/views/1?pane=issue&itemId=92306093&issue=ita-social-projects%7CGreenCity%7C7970)
Changes:

Add different profile privacy states like: PRIVATE(visible only for user), FRIENDS_ONLY(visible for user and user and their friends), PRIVATE(visible for all users)
Create new Liquibase changelog to change column type in DB and update previous values

Now Response looks like this:
Screenshot from 2025-01-06 21-26-33

### DIFF
--- a/service-api/src/main/java/greencity/enums/ProfilePrivacyPolicy.java
+++ b/service-api/src/main/java/greencity/enums/ProfilePrivacyPolicy.java
@@ -1,6 +1,6 @@
 package greencity.enums;
 
-public enum EcoPlacePrivacyPolicy {
+public enum ProfilePrivacyPolicy {
     PRIVATE,
     FRIENDS_ONLY,
     PUBLIC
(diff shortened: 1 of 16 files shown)


KizerovDmitriy (approved):

service-api/src/main/java/greencity/enums/ProfilePrivacyPolicy.java
```diff
@@ -0,0 +1,7 @@
+package greencity.enums;
+
+public enum ProfilePrivacyPolicy {
```
RomaMocherniuk, line 3: Why do we need two identical enums?

Cr1stal423: Because these enums are in different maven modules

Warded120 (approved):

RomaMocherniuk (approved):

softservedata (approved):

Warded120 (approved):
is it possible to change the names of fields (showLocation, showEcoPlace, showShoppingList) to be more intuitive, according to enum name (ProfilePrivacyPolicy)? e.g. locationPrivacyPolicy, ecoPlacesPrivacyPolicy and shoppingListPrivacyPolicy?

Issue #7988, opened 2025-01-05
Synchronize places from database with google places API


Warded120:
Implement functionality to synchronize places between greencity database and Google Places API.

- add functionality to fetch\filter places from Google Places API

Acceptance Criteria:

- All places retrieved from Google Places API are processed in the same way as places from database
- New places created in the system are shown on the Google Places map.
- Duplicate entries are avoided during synchronization.
- Synchronized places are displayed correctly in the frontend.

PR #8026, opened 2025-01-14
Synchronize places from database with google places api

Warded120:
Issue Link: https://github.com/ita-social-projects/GreenCity/issues/7988
--PROD--

Changes:

- Duplicates are avoided when adding places from ui to database
- Created google api service method to fetch from API
- Created endpoint to fetch places from places API

## Release Notes

- **New Features**
	- Added a new endpoint for filtering places using external Google Places API.
	- Introduced place filtering capabilities with advanced search parameters.
	- Enhanced location and place management with duplicate prevention.
	- New error messages for location validation and existing places.

- **Bug Fixes**
	- Improved error handling for existing place coordinates.
	- Added validation to prevent duplicate place entries.

- **Documentation**
	- Updated API documentation for new place filtering endpoint.

- **Chores**
	- Integrated Google Maps Services dependency.
	- Added new data transfer objects and service methods for API interactions.

### DIFF
--- a/core/src/main/java/greencity/controller/PlaceController.java
+++ b/core/src/main/java/greencity/controller/PlaceController.java
@@ -1,10 +1,12 @@
 package greencity.controller;
 
 import greencity.annotations.ApiPageable;
+import greencity.annotations.CurrentUser;
 import greencity.constant.HttpStatuses;
 import greencity.dto.PageableDto;
 import greencity.dto.favoriteplace.FavoritePlaceDto;
 import greencity.dto.filter.FilterPlaceDto;
+import greencity.dto.filter.FilterPlacesApiDto;
 import greencity.dto.place.AddPlaceDto;
 import greencity.dto.place.AdminPlaceDto;
 import greencity.dto.place.BulkUpdatePlaceStatusDto;
@@ -225,6 +227,30 @@ public class PlaceController {
         @Valid @RequestBody FilterPlaceDto filterDto) {
         return ResponseEntity.status(HttpStatus.OK)
             .body(placeService.getPlacesByFilter(filterDto));
+    }
+
+    /**
+     * The method which return a list of {@link PlaceByBoundsDto} filtered by values
+     * contained in the incoming {@link FilterPlacesApiDto} object.
+     * {@link PlaceByBoundsDto} are retrieved from Google Places API
+     *
+     * @param filterDto contains all information about the filtering of the list.
+     * @return a list of {@code PlaceByBoundsDto}
+     */
+    @ApiOperation(value = "Return a list places from Google Geocoding API filtered by values contained "
+            + "in the incoming FilterPlaceDto object")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = HttpStatuses.OK, response = FilterPlacesApiDto.class),
+            @ApiResponse(code = 303, message = HttpStatuses.SEE_OTHER),
+            @ApiResponse(code = 400, message = HttpStatuses.BAD_REQUEST),
+            @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
+            @ApiResponse(code = 404, message = HttpStatuses.NOT_FOUND)
+    })
+    @PostMapping("/filter/api")
+    public ResponseEntity<List<PlaceByBoundsDto>> getFilteredPlacesFromApi(
+            @Valid @RequestBody FilterPlacesApiDto filterDto,
+            @ApiIgnore @CurrentUser UserVO userVO) {
+        return ResponseEntity.ok().body(placeService.getPlacesByFilter(filterDto, userVO));
     }
 
     /**
--- a/core/src/main/java/greencity/controller/PlaceController.java
+++ b/core/src/main/java/greencity/controller/PlaceController.java
@@ -241,9 +241,7 @@ public class PlaceController {
         + "in the incoming FilterPlaceDto object")
     @ApiResponses(value = {
         @ApiResponse(code = 200, message = HttpStatuses.OK, response = FilterPlacesApiDto.class),
-        @ApiResponse(code = 303, message = HttpStatuses.SEE_OTHER),
         @ApiResponse(code = 400, message = HttpStatuses.BAD_REQUEST),
-        @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
         @ApiResponse(code = 404, message = HttpStatuses.NOT_FOUND)
     })
     @PostMapping("/filter/api")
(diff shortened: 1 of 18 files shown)


core/src/main/java/greencity/controller/PlaceController.java
```diff
@@ -227,6 +229,30 @@ public ResponseEntity<List<PlaceByBoundsDto>> getFilteredPlaces(
...
+     * {@link PlaceByBoundsDto} are retrieved from Google Places API
+     *
+     * @param filterDto contains all information about the filtering of the list.
+     * @return a list of {@code PlaceByBoundsDto}
+     */
+    @ApiOperation(value = "Return a list places from Google Geocoding API filtered by values contained "
+        + "in the incoming FilterPlaceDto object")
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = HttpStatuses.OK, response = FilterPlacesApiDto.class),
+        @ApiResponse(code = 303, message = HttpStatuses.SEE_OTHER),
```
Cr1stal423: In what cases can 303 response code be used? And whether such a response can be found here at all

service/src/main/java/greencity/service/GoogleApiService.java
```diff
@@ -54,6 +60,90 @@ public List<GeocodingResult> getResultFromGeoCode(String searchRequest) {
...
+
+    /**
+     * Parses a location string (latitude,longitude) into a LatLng object.
+     *
+     * @param location - A string in the format "latitude,longitude".
+     * @return LatLng object representing the location.
+     */
+    private com.google.maps.model.LatLng getLocationFromUserVO(UserVO userVO) {
+        return new com.google.maps.model.LatLng(userVO.getUserLocationDto().getLatitude(),
+            userVO.getUserLocationDto().getLongitude());
```
Cr1stal423, line 151: Potentially NullPointerException as coderabit wrote

core/src/main/java/greencity/controller/PlaceController.java
```diff
@@ -227,6 +229,30 @@ public ResponseEntity<List<PlaceByBoundsDto>> getFilteredPlaces(
...
+        @ApiResponse(code = 400, message = HttpStatuses.BAD_REQUEST),
+        @ApiResponse(code = 403, message = HttpStatuses.FORBIDDEN),
```
RomaMocherniuk: How can we have `403` response?
Based on the config this endpoint is available for the next roles `USER, ADMIN, MODERATOR, UBS_EMPLOYEE`
Do we have some roles that do not have access to this endpoint?

service-api/src/main/java/greencity/dto/filter/FilterPlacesApiDto.java
```diff
@@ -0,0 +1,26 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class FilterPlacesApiDto {
```
RomaMocherniuk, line 16: you can use `record` for it

service/src/main/java/greencity/service/GoogleApiService.java
```diff
@@ -54,6 +60,90 @@ public List<GeocodingResult> getResultFromGeoCode(String searchRequest) {
...
+     * Applies filters to a request from {@link FilterPlacesApiDto}.
+     *
+     * @param filterDto DTO containing filter parameters for the Places API request.
+     * @param request   API request filters are applied to
+     *
+     * @author Hrenevych Ivan
+     */
+    private static void applyFiltersToRequest(FilterPlacesApiDto filterDto, NearbySearchRequest request) {
+        if (filterDto.getKeyword() != null) {
+            request.keyword(filterDto.getKeyword());
```
RomaMocherniuk, line 109: what are the default values of all these fields(`keyword`, `type`,...) in the `NearbySearchRequest`?
If they are `null` we do not need these checkers

service/src/main/java/greencity/service/PlaceServiceImpl.java
```diff
@@ -570,6 +592,11 @@ public PlaceResponse addPlaceFromUi(AddPlaceDto dto, String email) {
             throw new UserBlockedException(ErrorMessage.USER_HAS_BLOCKED_STATUS);
         }
         List<GeocodingResult> geocodingResults = googleApiService.getResultFromGeoCode(dto.getLocationName());
+        double lat = geocodingResults.get(0).geometry.location.lat;
```
RomaMocherniuk, line 598: can we have `IndexOutOfBound` exception here?

service/src/main/java/greencity/service/GoogleApiService.java
Warded120: Params in a request are stored in a map. If we pass them, we set a key-value pair. they don't have a default value. So if a param (keyword, type etc.) is null, it is not passed.

service-api/src/main/java/greencity/dto/filter/FilterPlacesApiDto.java
Warded120: It's a PR for prod. Java 11 does not support records

core/src/main/java/greencity/controller/PlaceController.java
Warded120: there is no such response code in this endpoint, now it is removed

core/src/main/java/greencity/controller/PlaceController.java
Warded120: that's a good point, now it is removed

service/src/main/java/greencity/service/PlaceServiceImpl.java
Warded120: yes we can have such exception, now i've added a check to prevent this. it also resolves similar potential issue in initializeGeoCodingResults(geocodingResults) method

service/src/main/java/greencity/service/GoogleApiService.java
Warded120: fixed this. Now all objects that can potentially be null are checked before they are accessed.

softservedata (approved):

service-api/src/main/java/greencity/constant/ErrorMessage.java
```diff
@@ -58,6 +58,7 @@ public final class ErrorMessage {
         "Subscriber with this email address not found in the database.";
     public static final String INVALID_UNSUBSCRIBE_TOKEN = "Invalid unsubscribe token";
     public static final String LOCATION_NOT_FOUND_BY_ID = "The location does not exist by this id: ";
+    public static final String LOCATION_NOT_FOUND = "Location must be provided either in filterDto or userVO";
```
RomaMocherniuk: `filterDto` and `userVO` are the objects in our Application and Users should not know about them. this message does not useful for the users that use our app

RomaMocherniuk (approved):
One small comment

service-api/src/main/java/greencity/constant/ErrorMessage.java
Warded120: changed to more understandable message

Cr1stal423 (approved):

PR #8031, opened 2025-01-15
feat(webMvcConfig): Added custom argument resolver for PageableHandle…

### LINKED ISSUE #6389: [API][Friends List][get-friends-find all friends] There is response code 200 OK instead of 400 Bad Request when User send invalid value
**Reproducible:** always
**Build found:** 21.09.2023

**Preconditions**
User is logged in, token is got.

**Steps to reproduce**
1. Open [Swagger friend-controller](https://greencity.greencity.social/swagger-ui.html#/friend-controller).
2. Choose [get /friends](https://greencity.greencity.social/swagger-ui.html#/operations/friend-controller/findAllFriendsOfUserUsingGET)
 (Find all friends).
3. Check a response for get 400 with invalid value for size (e.g. 'jhnj') 

**Actual result**
The Status Code is 200 OK instead of 400 Bad Request.

**Expected result**
The Status Code is 400 Bad Request.

**User story and test case links**
E.g.: "User story #2122 
[Test case](https://jira.softserve.academy/browse/4184)"

Cr1stal423:
😎 GreenCity PR😎

Issue Link:
[6389](https://github.com/orgs/ita-social-projects/projects/43/views/1?filterQuery=assignee%3ACr1stal423&pane=issue&itemId=69330347&issue=ita-social-projects%7CGreenCity%7C6389)
Changes:

Add custom argument resolver for PageableHandlerMethodArgumentResolver
Add new error messages

Previous Condition:

image

Current Condition:

image

## Summary by CodeRabbit

- **New Features**
	- Enhanced pagination handling with a custom argument resolver.
	- Introduced constants for pagination defaults and limits.
	- Added specific error messages for size and page validation.

- **Bug Fixes**
	- Improved error handling for pagination parameters, ensuring better validation and messaging.

- **Tests**
	- Added comprehensive tests for the custom pagination argument resolver, covering normal and exceptional scenarios.

### DIFF
--- a/core/src/main/java/greencity/config/CustomPageableHandlerMethodArgumentResolver.java
+++ b/core/src/main/java/greencity/config/CustomPageableHandlerMethodArgumentResolver.java
@@ -21,16 +21,18 @@ public class CustomPageableHandlerMethodArgumentResolver extends PageableHandler
         ModelAndViewContainer mavContainer,
         NativeWebRequest webRequest,
         WebDataBinderFactory binderFactory) {
-        int page = parseParameter(webRequest.getParameter(PAGE), DEFAULT_PAGE, PAGE);
-        int size = parseParameter(webRequest.getParameter(SIZE), DEFAULT_PAGE_SIZE, SIZE);
+        int page = parseParameter(webRequest, PAGE, DEFAULT_PAGE);
+        int size = parseParameter(webRequest, SIZE, DEFAULT_PAGE_SIZE);
 
         if (size > MAX_PAGE_SIZE) {
             throw new BadRequestException(ErrorMessage.MAX_PAGE_SIZE_EXCEPTION);
         }
+
         return PageRequest.of(page, size);
     }
 
-    private int parseParameter(String paramValue, int defaultValue, String param) {
+    private int parseParameter(NativeWebRequest webRequest, String param, int defaultValue) {
+        String paramValue = webRequest.getParameter(param);
         if (paramValue == null) {
             return defaultValue;
         }
@@ -38,17 +40,12 @@ public class CustomPageableHandlerMethodArgumentResolver extends PageableHandler
             int value = Integer.parseInt(paramValue);
 
             if (value < 0) {
-                String errorMessage = param.equals(PAGE)
-                    ? ErrorMessage.NEGATIVE_PAGE_VALUE_EXCEPTION
-                    : ErrorMessage.NEGATIVE_SIZE_VALUE_EXCEPTION;
-                throw new IllegalArgumentException(errorMessage);
+                throw new IllegalArgumentException(String.format(ErrorMessage.NEGATIVE_VALUE_EXCEPTION, param));
             }
+
             return value;
         } catch (NumberFormatException e) {
-            String errorMessage = param.equals(PAGE)
-                ? ErrorMessage.INVALID_PAGE_VALUE_EXCEPTION
-                : ErrorMessage.INVALID_SIZE_VALUE_EXCEPTION;
-            throw new IllegalArgumentException(errorMessage, e);
+            throw new IllegalArgumentException(String.format(ErrorMessage.INVALID_VALUE_EXCEPTION, param), e);
         }
     }
 }
--- a/core/src/main/java/greencity/config/CustomPageableHandlerMethodArgumentResolver.java
+++ b/core/src/main/java/greencity/config/CustomPageableHandlerMethodArgumentResolver.java
@@ -1,0 +1,54 @@
+package greencity.config;
+
+import greencity.constant.ErrorMessage;
+import greencity.exception.exceptions.BadRequestException;
+import org.springframework.core.MethodParameter;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.web.PageableHandlerMethodArgumentResolver;
+import org.springframework.web.bind.support.WebDataBinderFactory;
+import org.springframework.web.context.request.NativeWebRequest;
+import org.springframework.web.method.support.ModelAndViewContainer;
+import static greencity.constant.PageableConstants.DEFAULT_PAGE;
+import static greencity.constant.PageableConstants.DEFAULT_PAGE_SIZE;
+import static greencity.constant.PageableConstants.MAX_PAGE_SIZE;
+import static greencity.constant.PageableConstants.PAGE;
+import static greencity.constant.PageableConstants.SIZE;
+
+public class CustomPageableHandlerMethodArgumentResolver extends PageableHandlerMethodArgumentResolver {
+    @Override
+    public Pageable resolveArgument(MethodParameter methodParameter,
+        ModelAndViewContainer mavContainer,
+        NativeWebRequest webRequest,
+        WebDataBinderFactory binderFactory) {
+        int page = parseParameter(webRequest.getParameter(PAGE), DEFAULT_PAGE, PAGE);
+        int size = parseParameter(webRequest.getParameter(SIZE), DEFAULT_PAGE_SIZE, SIZE);
+
+        if (size > MAX_PAGE_SIZE) {
+            throw new BadRequestException(ErrorMessage.MAX_PAGE_SIZE_EXCEPTION);
+        }
+        return PageRequest.of(page, size);
+    }
+
+    private int parseParameter(String paramValue, int defaultValue, String param) {
+        if (paramValue == null) {
+            return defaultValue;
+        }
+        try {
+            int value = Integer.parseInt(paramValue);
+
+            if (value < 0) {
+                String errorMessage = param.equals(PAGE)
+                    ? ErrorMessage.NEGATIVE_PAGE_VALUE_EXCEPTION
+                    : ErrorMessage.NEGATIVE_SIZE_VALUE_EXCEPTION;
+                throw new IllegalArgumentException(errorMessage);
+            }
+            return value;
+        } catch (NumberFormatException e) {
+            String errorMessage = param.equals(PAGE)
+                ? ErrorMessage.INVALID_PAGE_VALUE_EXCEPTION
+                : ErrorMessage.INVALID_SIZE_VALUE_EXCEPTION;
+            throw new IllegalArgumentException(errorMessage, e);
+        }
+    }
+}
--- a/core/src/main/java/greencity/config/WebMvcConfig.java
+++ b/core/src/main/java/greencity/config/WebMvcConfig.java
@@ -1,6 +1,5 @@
 package greencity.config;
 
-import greencity.constant.ErrorMessage;
 import greencity.converters.UserArgumentResolver;
 import greencity.service.UserService;
 import java.util.List;
@@ -11,9 +10,6 @@ import org.springframework.context.MessageSource;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.support.ReloadableResourceBundleMessageSource;
-import org.springframework.core.MethodParameter;
-import org.springframework.data.domain.PageRequest;
-import org.springframework.data.domain.Pageable;
 import org.springframework.data.web.PageableHandlerMethodArgumentResolver;
 import org.springframework.validation.beanvalidation.LocalValidatorFactoryBean;
 import org.springframework.web.method.support.HandlerMethodArgumentResolver;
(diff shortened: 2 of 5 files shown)


core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +101,35 @@ public void addInterceptors(InterceptorRegistry registry) {
 
     @Override
     public void addArgumentResolvers(List<HandlerMethodArgumentResolver> resolvers) {
+        resolvers.removeIf(resolver -> resolver instanceof PageableHandlerMethodArgumentResolver);
         resolvers.add(new UserArgumentResolver(userService, modelMapper));
+        resolvers.add(new PageableHandlerMethodArgumentResolver() {
+            @Override
+            public Pageable resolveArgument(MethodParameter methodParameter,
+                org.springframework.web.method.support.ModelAndViewContainer mavContainer,
```
KizerovDmitriy: Move this to import block

KizerovDmitriy (changes requested):

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: already refactored

softservedata (approved):

Maryna-511750 (approved):

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -58,7 +66,7 @@ public LocalValidatorFactoryBean getValidator() {
      * @return {@link SessionLocaleResolver}
      */
     @Bean
-    public LocaleResolver localeResolver() {
+    public LocaleResolver localeResolvers() {
```
RomaMocherniuk: why did you change the name to plural?

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +104,35 @@ public void addInterceptors(InterceptorRegistry registry) {
...
+        resolvers.removeIf(resolver -> resolver instanceof PageableHandlerMethodArgumentResolver);
         resolvers.add(new UserArgumentResolver(userService, modelMapper));
+        resolvers.add(new PageableHandlerMethodArgumentResolver() {
+            @Override
+            public Pageable resolveArgument(MethodParameter methodParameter,
+                ModelAndViewContainer mavContainer,
+                NativeWebRequest webRequest,
+                WebDataBinderFactory binderFactory) {
+                String pageParam = webRequest.getParameter("page");
+                String sizeParam = webRequest.getParameter("size");
```
RomaMocherniuk, line 116: ple

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +104,35 @@ public void addInterceptors(InterceptorRegistry registry) {
...
+
+                int page = 0;
+                int size = 20;
```
RomaMocherniuk: please move `page`, size, `0` and `20` to the constants

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +104,35 @@ public void addInterceptors(InterceptorRegistry registry) {
...
+
+                try {
+                    if (pageParam != null) {
+                        page = Integer.parseInt(pageParam);
```
RomaMocherniuk: should we check if the page value is not negative as well?

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +104,35 @@ public void addInterceptors(InterceptorRegistry registry) {
 
     @Override
     public void addArgumentResolvers(List<HandlerMethodArgumentResolver> resolvers) {
+        resolvers.removeIf(resolver -> resolver instanceof PageableHandlerMethodArgumentResolver);
         resolvers.add(new UserArgumentResolver(userService, modelMapper));
+        resolvers.add(new PageableHandlerMethodArgumentResolver() {
```
RomaMocherniuk: please move this logic to the separate class and add tests for it

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: Spring Context already had a bean with this name, so I'll think about how to name it differently

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: I will change it

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: Yes, I think also need to check for negative values, I'll add the following logic

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: Move the method override to another class?

core/src/main/java/greencity/config/WebMvcConfig.java
RomaMocherniuk: Usually, the plural is used for collections.
It is better to use an other name

RomaMocherniuk:
Please cover the new code with tests
Screenshot 2025-01-17 at 23 01 22

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +106,37 @@ public void addInterceptors(InterceptorRegistry registry) {
...
+                            throw new NumberFormatException(ErrorMessage.NEGATIVE_PAGE_VALUE_EXCEPTION);
+                        }
+                        page = Integer.parseInt(pageParam);
+                    }
+                    if (sizeParam != null) {
+                        if (!sizeParam.matches("\\d+")) {
+                            throw new NumberFormatException(ErrorMessage.NEGATIVE_SIZE_VALUE_EXCEPTION);
+                        }
+                        size = Integer.parseInt(sizeParam);
+                    }
```
RomaMocherniuk: you have code duplication
please move it to the method

core/src/main/java/greencity/config/WebMvcConfig.java
RomaMocherniuk: I see that you create a new class here and override some methods
```
new PageableHandlerMethodArgumentResolver() {
@Override
            public Pageable resolveArgument(...)
            ...
 }
```
Please move it to the separate class ad cover the new logic with tests

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +106,37 @@ public void addInterceptors(InterceptorRegistry registry) {
...
+                } catch (NumberFormatException ex) {
+                    throw new IllegalArgumentException(ErrorMessage.INVALID_SIZE_VALUE_EXCEPTION);
```
RomaMocherniuk: we can throw this exception if an invalid page as well, not only an invalid size

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +106,37 @@ public void addInterceptors(InterceptorRegistry registry) {
...
+                        if (!pageParam.matches("\\d+")) {
+                            throw new NumberFormatException(ErrorMessage.NEGATIVE_PAGE_VALUE_EXCEPTION);
+                        }
+                        page = Integer.parseInt(pageParam);
+                    }
+                    if (sizeParam != null) {
+                        if (!sizeParam.matches("\\d+")) {
+                            throw new NumberFormatException(ErrorMessage.NEGATIVE_SIZE_VALUE_EXCEPTION);
+                        }
+                        size = Integer.parseInt(sizeParam);
```
RomaMocherniuk: do we have some upper limit for the size?
What would happen if I pass size=1000000
Will we return all records from DB?
We might have a performance problem in this case if there are a lot of records in the DB

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +106,37 @@ public void addInterceptors(InterceptorRegistry registry) {
 
     @Override
     public void addArgumentResolvers(List<HandlerMethodArgumentResolver> resolvers) {
+        resolvers.removeIf(resolver -> resolver instanceof PageableHandlerMethodArgumentResolver);
```
RomaMocherniuk, line 100: why do you remove `PageableHandlerMethodArgumentResolvers` here?

core/src/main/java/greencity/config/WebMvcConfig.java
RomaMocherniuk: I wonder if we really need to throw an exception here?
It is better to use default values

core/src/main/java/greencity/config/WebMvcConfig.java
```diff
@@ -96,6 +106,37 @@ public void addInterceptors(InterceptorRegistry registry) {
...
         resolvers.add(new UserArgumentResolver(userService, modelMapper));
+        resolvers.add(new PageableHandlerMethodArgumentResolver() {
+            @Override
+            public Pageable resolveArgument(MethodParameter methodParameter,
+                ModelAndViewContainer mavContainer,
+                NativeWebRequest webRequest,
+                WebDataBinderFactory binderFactory) {
+                String pageParam = webRequest.getParameter("page");
+                String sizeParam = webRequest.getParameter("size");
```
RomaMocherniuk: please move the stings `page` and `size` to the constants too

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: Okay, I'm gonna put it in a separate method.

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: This is the core  of the task: if the size field is entered incorrectly, adjust the setting to the default value and return a response about the incorrectly entered field

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: In this way, I exclude the default implementation and overwrite it with my own, but maybe there is another way to do it?

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: I missed it, so it should also be put in constants

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: As far as I know, the default implementation of Pageable in Spring doesn't have a specific limit either, but that's a good point, I'll add a limit

core/src/main/java/greencity/config/WebMvcConfig.java
Cr1stal423: I have completely changed everything according to the comments, please review again

core/src/main/java/greencity/config/CustomPageableHandlerMethodArgumentResolver.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public Pageable resolveArgument(MethodParameter methodParameter,
+        ModelAndViewContainer mavContainer,
+        NativeWebRequest webRequest,
+        WebDataBinderFactory binderFactory) {
+        int page = parseParameter(webRequest.getParameter(PAGE), DEFAULT_PAGE, PAGE);
+        int size = parseParameter(webRequest.getParameter(SIZE), DEFAULT_PAGE_SIZE, SIZE);
+
+        if (size > MAX_PAGE_SIZE) {
+            throw new BadRequestException(ErrorMessage.MAX_PAGE_SIZE_EXCEPTION);
```
RomaMocherniuk, line 28: it is better to use the default page size instead of throwing Exception. But it depends on the requirements. If you have to throw an exception it is ok

core/src/main/java/greencity/config/CustomPageableHandlerMethodArgumentResolver.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+        return PageRequest.of(page, size);
+    }
+
+    private int parseParameter(String paramValue, int defaultValue, String param) {
```
RomaMocherniuk: can be optimized
```
 private int parseParameter(NativeWebRequest webRequest, String param, int defaultValue) {
        String paramValue = webRequest.getParameter(param);
        if (paramValue == null) {
            return defaultValue;
        }
        try {
            int value = Integer.parseInt(paramValue);

            if (value < 0) {
                throw new IllegalArgumentException(String.format(NEGATIVE_VALUE_EXCEPTION, param));
            }

            return value;
        } catch (NumberFormatException e) {
            throw new IllegalArgumentException(String.format(INVALID_VALUE_EXCEPTION, param), e);
        }
    }

    public static final String INVALID_VALUE_EXCEPTION = "Invalid value for %s: must be an integer";
    public static final String NEGATIVE_VALUE_EXCEPTION = "%s must be a positive number";
```

And usage of this method

```
@Override
    public Pageable resolveArgument(MethodParameter methodParameter,
                                     ModelAndViewContainer mavContainer,
                                     NativeWebRequest webRequest,
                                     WebDataBinderFactory binderFactory) {
        int page = parseParameter(webRequest, PAGE, DEFAULT_PAGE);
        int size = parseParameter(webRequest, SIZE, DEFAULT_PAGE_SIZE);

        if (size > MAX_PAGE_SIZE) {
             throw new BadRequestException(ErrorMessage.MAX_PAGE_SIZE_EXCEPTION);
        }

        return PageRequest.of(page, size);
    }
```

core/src/main/java/greencity/config/CustomPageableHandlerMethodArgumentResolver.java
Cr1stal423: It will not be very clear to the user if he form a size value 10000 and is returned a value of 20

RomaMocherniuk (approved):

KizerovDmitriy (approved):

PR #8042, opened 2025-01-17
Incorrect 404 Not Found Response in Swagger Documentation

Cr1stal423:
😎 GreenCity PR😎

Issue Link:
[8017](https://github.com/orgs/ita-social-projects/projects/43/views/1?filterQuery=assignee%3ACr1stal423&pane=issue&itemId=93483112&issue=ita-social-projects|GreenCity|8017)

Changes:

Add proper exception handling for specific endpoints
Add new error messages

Previous Condition:

image

Current Condition:

image

## Summary by CodeRabbit

- **API Documentation**
	- Removed "404 Not Found" response codes from friend-related API methods.
	- Improved error handling for unsupported HTTP methods.

- **Configuration**
	- Added configuration for valid API endpoints and statuses.

- **Error Handling**
	- Introduced more detailed error responses for unsupported HTTP methods.
	- Added new error messages for various HTTP-related scenarios.

- **New Features**
	- Added new classes for endpoint validation and response building.
	- Enhanced validation logic for API endpoints.

- **Tests**
	- Introduced comprehensive unit tests for endpoint validation and error response handling.
	- Added tests for the new error response builder functionality.
	- Added tests for the `EndpointValidator` functionality.
	- Updated test setups to accommodate changes in exception handling.


core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -51,19 +54,26 @@
...
  * Custom exception handler.
  */
-@AllArgsConstructor
 @RestControllerAdvice
 @Slf4j
 public class CustomExceptionHandler extends ResponseEntityExceptionHandler {
     private ErrorAttributes errorAttributes;
     private final ObjectMapper objectMapper;
+    @Value("${valid.endpoints}")
+    private List<String> validEndpoints;
```
RomaMocherniuk: The exception handler is responsible for handling exceptions only
Please add sone validation filters for this logc

core/src/main/resources/application.properties
```diff
@@ -42,4 +42,5 @@ spring.messaging.stomp.websocket.allowed-origins=\
   http://localhost:4200/, \
   http://localhost:4205/*
 
-server.tomcat.relaxed-query-chars=<, >, [, \,, ], ^, `, {, |, }
\ No newline at end of file
+server.tomcat.relaxed-query-chars=<, >, [, \,, ], ^, `, {, |, }
+valid.endpoints=/friends/not-friends-yet,/friends/friendRequests
```
RomaMocherniuk: Do we have only 2 valid endpoints in the whole application?

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -558,4 +567,69 @@ public final ResponseEntity<Object> handleResourceNotFoundException(ResourceNotF
...
+    protected ResponseEntity<Object> handleHttpRequestMethodNotSupported(
+        HttpRequestMethodNotSupportedException ex, HttpHeaders headers, HttpStatusCode status, WebRequest request) {
+        String url = getUrlFromRequest(request);
+        ServletWebRequest servletWebRequest = (ServletWebRequest) request;
+        HttpServletRequest servletRequest = servletWebRequest.getRequest();
+
+        String method = servletRequest.getMethod();
+        List<String> list = headers.getOrEmpty("Allow");
+
+        switch (getCondition(url, method, list, validEndpoints)) {
```
RomaMocherniuk: Is there any reason to pass `validEndpoints` as a parameter?
we have access to it in `getCondition` method

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -558,4 +567,69 @@ public final ResponseEntity<Object> handleResourceNotFoundException(ResourceNotF
...
+        }
+    }
+
+    private boolean hasExtraCharacters(String url, List<String> validEndpoints) {
+        boolean isValidEndpoint = validEndpoints.contains(url);
+        if (!isValidEndpoint) {
+            return true;
+        }
+        return false;
+    }
```
RomaMocherniuk: can be simplified `return !validEndpoints.contains(url);`
And this method can be removed at all

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -558,4 +567,69 @@ public final ResponseEntity<Object> handleResourceNotFoundException(ResourceNotF
...
+
+    private String getUrlFromRequest(WebRequest request) {
+        return request.getDescription(false).replace("uri=", "");
```
RomaMocherniuk: why don't you use `request.getUri()` ?

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -558,4 +567,69 @@ public final ResponseEntity<Object> handleResourceNotFoundException(ResourceNotF
...
+    }
+
+    private String getCondition(String url, String method, List<String> list, List<String> validEndpoints) {
+        if (hasExtraCharacters(url, validEndpoints)) {
+            return "extraCharacters";
+        } else if (!list.contains(method)) {
+            return "methodNotAllowed";
+        }
+        return "default";
```
RomaMocherniuk: please move all the sting you use to the constants

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -558,4 +567,69 @@ public final ResponseEntity<Object> handleResourceNotFoundException(ResourceNotF
...
+    }
+
+    private Map<String, Object> setResponse(HttpStatus httpStatus, String error, String errorMessage, String url) {
```
RomaMocherniuk: The method name is incorrect. Usually, `set` prefix is used for setters 
`buildResponse` Body  is better name for it

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -558,4 +567,69 @@ public final ResponseEntity<Object> handleResourceNotFoundException(ResourceNotF
...
+        String method = servletRequest.getMethod();
+        List<String> list = headers.getOrEmpty("Allow");
+
+        switch (getCondition(url, method, list, validEndpoints)) {
+            case "extraCharacters":
+                String notFoundErrorMessage = String.format("No endpoint found for %s", url);
+                Map<String, Object> body1 = setResponse(HttpStatus.NOT_FOUND, "Not Found", notFoundErrorMessage, url);
+                return ResponseEntity.status(HttpStatus.NOT_FOUND).body(body1);
+
+            case "methodNotAllowed":
```
RomaMocherniuk: `METHOD_NOT_ALLOWED` response code should be returned when the endpoint exists but a client uses an incorrect `HTTP` method. for example. we have some `POST` endpoint but client use GET method to call it

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -558,4 +567,69 @@ public final ResponseEntity<Object> handleResourceNotFoundException(ResourceNotF
...
+        ServletWebRequest servletWebRequest = (ServletWebRequest) request;
+        HttpServletRequest servletRequest = servletWebRequest.getRequest();
+
+        String method = servletRequest.getMethod();
+        List<String> list = headers.getOrEmpty("Allow");
+
+        switch (getCondition(url, method, list, validEndpoints)) {
+            case "extraCharacters":
+                String notFoundErrorMessage = String.format("No endpoint found for %s", url);
+                Map<String, Object> body1 = setResponse(HttpStatus.NOT_FOUND, "Not Found", notFoundErrorMessage, url);
```
RomaMocherniuk: `NOT_FOUND` in general is returned when a client asks for something but this resource does not exist
For example, we want to get info about user with `id=123` but there is not user with such id

core/src/main/resources/application.properties
Cr1stal423: I added the endpoints that are relevant to my task, but since this method will globally define the handling of HttpRequestMethodNotSupported, it would be better to add all endpoints, I will do it

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
Cr1stal423: Yes, it makes no sense to pass the parameter, I'll fix it

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
Cr1stal423: However, this is a way to simplify the code, I will change it

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
Cr1stal423: There is no such method in WebRequest

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
Cr1stal423: This is a good decision, I will change it

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
Cr1stal423: I'll change the method name

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
Cr1stal423: The task specified that it was necessary to receive a 404 response when the resource was not found. "Investigate and correct the API behavior to return 404 Not Found for non-existent resources"

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
Cr1stal423: So in this case METHOD_NOT_ALLOWED and is used in this case, the method whose default implementation I rewrote is responsible for this logic and in my new logic it also checks whether the method is in the list of available methods, or did I understand something wrong?

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final String METHOD_NOT_ALLOWED = "methodNotAllowed";
+    private static final String EXTRA_CHARACTERS = "extraCharacters";
+
+    private EndpointValidationHelper() {
+    }
+
+    public static ResponseEntity<Object> response(HttpRequestMethodNotSupportedException ex, HttpHeaders headers,
+        WebRequest request) {
+        String url = getUrlFromRequest(request);
+        ServletWebRequest servletWebRequest = (ServletWebRequest) request;
```
RomaMocherniuk: Can we have `ClassCastExceprion` here?
Please add `instanceOf`

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+        HttpServletRequest servletRequest = servletWebRequest.getRequest();
+
+        String method = servletRequest.getMethod();
+        List<String> list = headers.getOrEmpty("Allow");
```
RomaMocherniuk: please move `Allow` to the constant and use better name for the variable

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+
+        switch (getCondition(url, method, list)) {
+            case EXTRA_CHARACTERS:
+                String notFoundErrorMessage = String.format("No endpoint found for %s", url);
+                return ExceptionResponseBuilder.buildResponse(HttpStatus.NOT_FOUND, "Not Found",
+                    notFoundErrorMessage, url);
```
RomaMocherniuk, line 43: please move all strings you use to the constants

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+public class EndpointValidationHelper {
+    private static final String METHOD_NOT_ALLOWED = "methodNotAllowed";
+    private static final String EXTRA_CHARACTERS = "extraCharacters";
+
+    private EndpointValidationHelper() {
+    }
+
+    public static ResponseEntity<Object> response(HttpRequestMethodNotSupportedException ex, HttpHeaders headers,
+        WebRequest request) {
+        String url = getUrlFromRequest(request);
```
RomaMocherniuk, line 28: you can move it after casting a request to servletWebRequest and get the URL like this
```
servletWebRequest.getRequest().getRequestURL();
```

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+        ServletWebRequest servletWebRequest = (ServletWebRequest) request;
+        HttpServletRequest servletRequest = servletWebRequest.getRequest();
+
+        String method = servletRequest.getMethod();
+        List<String> list = headers.getOrEmpty("Allow");
+
+        switch (getCondition(url, method, list)) {
```
RomaMocherniuk: you can use 
```
return switch(getCondition(url, method, list)) {
...
}
```

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+        if (!list.contains(method)) {
+            return METHOD_NOT_ALLOWED;
+        }
+        return "default";
+    }
+
+    public static String getErrorMessage(HttpRequestMethodNotSupportedException ex, String url,
+        String supportedMethods) {
+        return String.format(
+            "Method %s is not allowed for %s. Supported Methods: %s",
```
RomaMocherniuk: please move strings to the constans

core/src/main/java/greencity/exception/helper/ExceptionResponseBuilder.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public static ResponseEntity<Object> buildResponse(HttpStatus httpStatus, String error, String errorMessage,
+        String url) {
+        Map<String, Object> response = new LinkedHashMap<>();
+        response.put("timestamp", System.currentTimeMillis());
+        response.put("status", httpStatus.value());
+        response.put("error", error);
+        response.put("message", errorMessage);
+        response.put("path", url);
```
RomaMocherniuk: the same here

core/src/main/java/greencity/validator/EndpointValidator.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private EndpointValidator() {
+    }
+
+    @Value("${valid.endpoints}")
+    public static void setValidEndpoints(List<String> validEndpoints) {
+        EndpointValidator.validEndpoints = validEndpoints;
+    }
+
+    public static boolean isValidEndpoint(String endpointTemplate, String actualUrl) {
```
RomaMocherniuk: can be private

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+            default:
+                return null;
+        }
+    }
+
+    public static String getUrlFromRequest(WebRequest request) {
+        return request.getDescription(false).replace("uri=", "");
+    }
+
+    public static String getCondition(String url, String method, List<String> list) {
```
RomaMocherniuk: please rename this method

core/src/main/java/greencity/validator/EndpointValidator.java
```diff
@@ -0,0 +1,63 @@
...
+            } else {
+                if (!templatePart.equals(urlPart)) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    public static boolean hasExtraCharacters(String url) {
```
RomaMocherniuk: can be private

core/src/main/resources/application.properties
```diff
@@ -42,4 +42,60 @@ spring.messaging.stomp.websocket.allowed-origins=\
   http://localhost:4200/, \
   http://localhost:4205/*
 
-server.tomcat.relaxed-query-chars=<, >, [, \,, ], ^, `, {, |, }
\ No newline at end of file
+server.tomcat.relaxed-query-chars=<, >, [, \,, ], ^, `, {, |, }
+valid.endpoints=/friends/not-friends-yet,/friends/friendRequests, friends/{friendId}, /ai/forecast/, /ai/generate/eco-news, /management/, \
```
RomaMocherniuk: please add to the Readme that in case we want to add a new endpoint this value should be updated

core/src/test/java/greencity/exception/handler/helper/EndpointValidatorHelperTest.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.Map;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.when;
+import static org.powermock.api.mockito.PowerMockito.mock;
+
+@ExtendWith(MockitoExtension.class)
+class EndpointValidatorHelperTest {
```
RomaMocherniuk, line 28: please add tests if we have placeholder in the `URL`, like `/management/users/{id}/delete-filter,  `/management/users/{id}/friends`, /place/{status}`

core/src/main/resources/application.properties
RomaMocherniuk: its a breaking  changes
please check `ALL` endpoints we have to be sure everything works fine

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
Cr1stal423: As far as I know, in this case there will be no such problem, because the request in this case will always be of type ServletWebRequest, but it's still a good idea to process it

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
Cr1stal423: Ок

core/src/main/resources/application.properties
Cr1stal423: Adding all the info in the readme about the need to add endpoints to valid.endpoints is a good point, I'll do that. For the 2nd comment, you mean check all endpoints in postman/swagger?

core/src/test/java/greencity/exception/handler/helper/EndpointValidatorHelperTest.java
Cr1stal423: I agree, we need placeholde to check such cases as well

core/src/main/java/greencity/validator/EndpointValidator.java
Cr1stal423: If the method is private, I will not test it, or I will use Reflection, which is not a very good practice, are there any potential problems in your opinion with public methods in this case?

core/src/main/java/greencity/exception/helper/ExceptionResponseBuilder.java
Cr1stal423: I'll fix it

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
Cr1stal423: I'll fix it

core/src/main/java/greencity/validator/EndpointValidator.java
RomaMocherniuk: you can test it using the public method where it is used

core/src/main/resources/application.properties
RomaMocherniuk: Just add to the readme that if we add a new endpoint it also should be added to this config file

core/src/main/java/greencity/exception/handler/CustomExceptionHandler.java
```diff
@@ -558,4 +562,14 @@ public final ResponseEntity<Object> handleResourceNotFoundException(ResourceNotF
...
     }
+
+    @Override
+    protected ResponseEntity<Object> handleHttpRequestMethodNotSupported(
+        HttpRequestMethodNotSupportedException ex, HttpHeaders headers, HttpStatusCode status, WebRequest request) {
+        ResponseEntity<Object> response = EndpointValidationHelper.response(ex, headers, request);
+        if (response == null) {
+            return super.handleHttpRequestMethodNotSupported(ex, headers, status, request);
+        }
+        return EndpointValidationHelper.response(ex, headers, request);
```
RomaMocherniuk: any reason to call `EndpointValidationHelper.response(ex, headers, request)` twice?
Why cant we return the `response` instead?

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
RomaMocherniuk: please move it to the` if block`

```
 String method;
 String URL;

        if (request instanceof ServletWebRequest servletWebRequest) {
            HttpServletRequest servletRequest = servletWebRequest.getRequest();
            method = servletRequest.getMethod();
            url = getUrlFromRequest(request);
        } else {
            throw new IllegalArgumentException(ErrorMessage.INVALID_REQUEST_TYPE);
        }
```

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
```diff
@@ -0,0 +1,73 @@
...
+        if (!EndpointValidator.checkUrl(url)) {
+            return EXTRA_CHARACTERS;
+        }
+        if (!allowedMethod.contains(method)) {
+            return METHOD_NOT_ALLOWED;
+        }
+        return DEFAULT_CONDITION;
+    }
+
+    public static String getErrorMessage(HttpRequestMethodNotSupportedException ex, String url,
```
RomaMocherniuk: I do not see any reason to create this method

core/src/main/java/greencity/validator/EndpointValidator.java
```diff
@@ -0,0 +1,66 @@
...
+
+import lombok.Getter;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import java.util.List;
+
+@Component
+public class EndpointValidator {
+    @Getter
+    private static List<String> validEndpoints = List.of();
```
RomaMocherniuk: do you use this variable somewhere else?
If yes please remove unmutable copy of this list to prevent modification

softservedata (approved):

core/src/main/java/greencity/exception/helper/EndpointValidationHelper.java
Cr1stal423: I agree, I've removed this method

core/src/main/java/greencity/validator/EndpointValidator.java
Cr1stal423: Removed the getter from this variable, it is not used anywhere outside the class

Warded120 (approved):

softservedata (approved):

core/src/main/resources/application.properties
```diff
@@ -42,4 +42,64 @@ spring.messaging.stomp.websocket.allowed-origins=\
...
+  /management/user/statistics/locations, /management/user/statistics/preferences, /notifications, /notifications/{notificationId}/viewNotification, /notifications/{notificationId}/unreadNotification \
+  /notifications/{notificationId, /notifications/notifications, /place/{placeId}/comments, comments/{id}, comments, /place/propose, /place/update, /place/info/{id}, /place/info/favorite/{placeId} \
+  /place/save/favorite/, /place/getListPlaceLocationByMapsBounds, /place/{status}, /place/filter, /place/status, /place/filter/predicate, /place/about/{id}, /place/statuses \
+  /place/{id}, /place, /place/v2/filteredPlacesCategories, /place/v2/save, /place/all, /chat, /search/eco-news, /search/events, /search/places, /social-networks/image, /social-networks, /specification \
+  /subscriptions, /subscriptions/{unsubscribeToken}, /tags/search, /tags/v2/search, /user/to-do-list-items/habits/{habitId}/to-do-list, /user/to-do-list-items, /user/to-do-list-items/{userToDoListItemId} \
+  /user/to-do-list-items/{userToDoListItemId}/status/{status}, /user/to-do-list-items/user-to-do-list-items, /user/to-do-list-items/{userId}/get-all-inprogress, /token \
+  /ai/forecast
+
```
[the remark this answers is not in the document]
Cr1stal423: fixed

maks741 (approved):

Issue #8043, opened 2025-01-20
[GreenCity. Friends]  Wrong order of users returned 


KostashchukIryna:
**Environment:** Swagger
**Reproducible:** always
**Build found:** "commitHash": "fcf05c3e8971c95e86bd53e7d34af258c8adb88d",
  "commitDate": "17/01/2025 11:39:23"

**Preconditions**
There are some users in the database.
The user is logged in, and the token is gotten.

**Steps to reproduce**
1. Go to https://greencity.greencity.cx.ua/swagger-ui/index.html#/friend-controller/findRecommendedFriends.
2. Type in sort: "id,asc" (or other criteria like name, email, exc.).
3. Click on "Execute".

**Actual result**
Users in response body are not sorted.

**Expected result**
Users are in order specified in "sort" parameter.

**User story and test case links**
#1444

**Labels to be added**
"Bug",  "API", "Friends"

LatyshevaBaranovska:
Sorting by "City" is not supported.

![Image](https://github.com/user-attachments/assets/7c61a529-98a3-4a5f-b10d-dc6c5d4b988e)
![Image](https://github.com/user-attachments/assets/2c650191-27cf-4d0a-bd4e-73dbc417fc57)
![Image](https://github.com/user-attachments/assets/dc5cec5e-7daf-4a77-afc9-637794a8df2e)
![Image](https://github.com/user-attachments/assets/ea9aa573-51d0-41c4-a191-5505731676a4)

PR #8048, opened 2025-01-20
Fixed sorting of users in friends service

KostashchukIryna:
### Issue Link 📋
#8043
### Issue summary
After calling fillListOfUserWithCountOfMutualFriendsAndChatIdForCurrentUser() previous order of users was lost.

## Summary by CodeRabbit

- **New Features**
	- Enhanced user mutual friends retrieval method with improved query handling and result processing.

- **Tests**
	- Added a new test to verify preservation of user order when retrieving mutual friends.
	- Modified existing test to ensure no query is created when the user list is empty.

### DIFF
--- a/dao/src/main/java/greencity/repository/impl/CustomUserRepoImpl.java
+++ b/dao/src/main/java/greencity/repository/impl/CustomUserRepoImpl.java
@@ -26,17 +26,17 @@ public class CustomUserRepoImpl implements CustomUserRepo {
         List<User> users) {
         Objects.requireNonNull(users);
 
+        if (users.isEmpty()) {
+            return Collections.emptyList();
+        }
+
         TypedQuery<UserFriendDto> query = entityManager
             .createNamedQuery("User.fillListOfUserWithCountOfMutualFriendsAndChatIdForCurrentUser",
                 UserFriendDto.class);
         query.setParameter("userId", userId);
 
-        List<Long> userIds;
-        if (users.isEmpty()) {
-            userIds = Collections.singletonList(-1L);
-        } else {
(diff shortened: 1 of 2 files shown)


dao/src/main/java/greencity/repository/impl/CustomUserRepoImpl.java
```diff
@@ -29,12 +30,22 @@ public List<UserFriendDto> fillListOfUserWithCountOfMutualFriendsAndChatIdForCur
...
+            userIds = Collections.singletonList(-1L);
         } else {
-            List<Long> userIds = users.stream().map(User::getId).collect(Collectors.toList());
-            query.setParameter("users", userIds);
+            userIds = users.stream().map(User::getId).collect(Collectors.toList());
         }
-        return query.getResultList();
+        query.setParameter("users", userIds);
+
+        List<UserFriendDto> resultList = query.getResultList();
```
RomaMocherniuk, line 42: Do we have any result if `userIds = Collections.singletonList(-1L);`?

KostashchukIryna: Query returns empty list in that case, so it would be better to return early and not call the db. Thanks for pointing out

KostashchukIryna: changed it

RomaMocherniuk (approved):

KizerovDmitriy (approved):

VasylyshynDmytro (approved):

urio999 (approved):

Issue #8047, opened 2025-01-20
[GreenCity. Friends] Sorting Parameters Not Working After Fix for Invalid Sort Parameters


KostashchukIryna:
**Environment:** Swagger
**Reproducible:** always
**Build found:** "commitHash": "2d94ac7fce116bc532e5f1dec5dd7def8bfa4859",
  "commitDate": "20/01/2025 15:27:26"

Sorting functionality no longer works after the fix implemented for #6382 (invalid sort parameters causing a 500 status code). Currently, any sorting parameter provided results in an UnsupportedSortException, making sorting unusable.

**Preconditions**
The user is logged in, and the token is gotten.

**Steps to reproduce**
1. Go to https://greencity.greencity.cx.ua/swagger-ui/index.html#/friend-controller/findAllUsersExceptMainUserAndUsersFriendAndRequestersToMainUser .
2. Send the request with valid sort parameters (for instance "id").
3. Observe the response.

**Actual result**
The status code is 400 Bad Request.
Any sorting parameter causes an exception (UnsupportedSortException: INVALID_SORTING_VALUE), even if the parameters are valid.

![Image](https://github.com/user-attachments/assets/52e08fc1-b662-4d89-865a-d78937b201af)

**Expected result**
The status code is 200 Success.

**User story and test case links**
#1003

**Labels to be added**
"Bug", "API"

PR #8072, opened 2025-01-27
Fixed validation of sort parameter in pageable

KostashchukIryna:
### Issue Link 📋
#8047

### Issue summary
 Previously, if an invalid sort field was provided, it would result in a SQL error, which would trigger a 500 status code. This approach could lead to unnecessary failures and poor user experience, especially when invalid sorting fields were provided. **Or** no sorting was allowed and anything with sort parameter would result in 400 code. **Or** SQL error was catched after executing wrong query which might be unnecessary.

### Solution
This PR proposes the validation of sorting parameters in the handler layer. 
Also considered using static registry of sortable values, using annotations and reflection or moving this logic to service layer.

## Release Notes

- **New Features**
	- Enhanced pagination and sorting capabilities across multiple controllers and services.
	- Introduced a new `Sortable` interface for defining sortable fields in DTOs.
	- Added robust validation for sorting parameters across different data transfer objects.

- **Improvements**
	- Improved API documentation by explicitly specifying DTO classes for pageable responses.
	- Implemented a centralized sorting validation mechanism.
	- Added support for sorting on various entity types including habits, friends, eco-news, and comments.

- **Technical Enhancements**
	- Created `CustomPageableHandlerMethodArgumentResolver` for advanced pagination handling.
	- Introduced `SortPageableValidator` to enforce sorting rules.
	- Added `SortableFields` enum to standardize sortable field definitions.

- **Bug Fixes**
	- Resolved inconsistencies in sorting and pagination across different controllers.


Maryna-511750 (approved):

core/src/main/java/greencity/handler/CustomPageableHandlerMethodArgumentResolver.java
```diff
@@ -16,10 +18,17 @@
...
 public class CustomPageableHandlerMethodArgumentResolver extends PageableHandlerMethodArgumentResolver {
+    private final CustomSortHandlerMethodArgumentResolver customSortResolver;
+
+    public CustomPageableHandlerMethodArgumentResolver(CustomSortHandlerMethodArgumentResolver customSortResolver) {
+        this.customSortResolver = customSortResolver;
+    }
+
     @Override
-    public Pageable resolveArgument(MethodParameter methodParameter,
+    @Nonnull
```
ChernenkoVitaliy, line 28: I don't know if it matters, but here should be `@NonNull` annotation from `org.springframework.lang.NonNull`;

core/src/main/java/greencity/handler/CustomPageableHandlerMethodArgumentResolver.java
```diff
@@ -16,10 +18,17 @@
...
+    public Pageable resolveArgument(@Nonnull MethodParameter methodParameter,
```
ChernenkoVitaliy, line 29: Same here. Looks like here should be `@Nullable` from `org.springframework.lang`.

core/src/main/java/greencity/handler/CustomSortHandlerMethodArgumentResolver.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.web.method.support.ModelAndViewContainer;
+
+/**
+ * Custom argument resolver for handling {@link Sort} parameters in API
+ * requests. This class extends the functionality of
+ * {@link SortHandlerMethodArgumentResolver} by adding validation for sort
+ * parameters based on the {@link ApiPageable} annotation.
+ *
+ */
+public class CustomSortHandlerMethodArgumentResolver implements SortArgumentResolver {
```
ChernenkoVitaliy, line 22: In my opinion, it would be much better to extend `SortHandlerMethodArgumentResolver` instead of implement  `SortArgumentResolver`. Then You don't need to pass `delegate` into the constructor and can use parent realization in your method. For example You can do next:
        `Sort sort = super.resolveArgument(parameter, mavContainer, webRequest, binderFactory);`

core/src/main/java/greencity/validator/SortPageableValidator.java
```diff
@@ -0,0 +1,56 @@
...
+        validFieldsMap.put(PlaceCommentAdminDto.class, new PlaceCommentAdminDto().getSortableFields());
+        validFieldsMap.put(RatingPointsDto.class, new RatingPointsDto().getSortableFields());
+        validFieldsMap.put(RatingStatisticsDtoForTables.class,
+            new RatingStatisticsDtoForTables().getSortableFields());
+        validFieldsMap.put(UserFriendDto.class, new UserFriendDto().getSortableFields());
+        return validFieldsMap;
+    }
+
+    public void validateSortParameters(Sort sort, Class<? extends Sortable> dtoClass) {
+        List<String> validFields = VALID_FIELDS_MAP.get(dtoClass);
```
ChernenkoVitaliy, line 43: For me, the method `initValidFieldsMap` can be a problem if somebody will try to add a new dto.
I mean, it can take some time to investigate the project to understand that here should be added a class in the map. It's not really obvious.

I'm not sure, but maybe it makes sense to use Reflection?
And do something like this instead:
```
            Sortable dto = dtoClass.getDeclaredConstructor().newInstance();
            List<String> validFields = dto.getSortableFields();
```

core/src/test/java/greencity/validator/SortPageableValidatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.data.domain.Sort;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class SortPageableValidatorTest {
+    private final SortPageableValidator sortPageableValidator = new SortPageableValidator();
+
+    @Test
+    void validateSortParametersWithValidSortField() {
```
ChernenkoVitaliy, line 19: According to the test convention used on the project, test names should end with the word '**Test**'.
https://github.com/ita-social-projects/GreenCity/wiki/Naming-Conventions

service-api/src/test/java/greencity/dto/econews/EcoNewsGenericDtoTest.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class EcoNewsGenericDtoTest {
+
+    @Test
+    void getSortableFields() {
```
ChernenkoVitaliy, line 14: Test names should end with the word '**Test**'.

service-api/src/test/java/greencity/dto/factoftheday/FactOfTheDayDTOTest.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class FactOfTheDayDTOTest {
+
+    @Test
+    void getSortableFields() {
```
ChernenkoVitaliy, line 14: The same here.

service-api/src/test/java/greencity/dto/habit/HabitDtoTest.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class HabitDtoTest {
+
+    @Test
+    void getSortableFields() {
```
ChernenkoVitaliy, line 14: The same.

service-api/src/test/java/greencity/dto/habit/HabitManagementDtoTest.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class HabitManagementDtoTest {
+
+    @Test
+    void getSortableFields() {
```
ChernenkoVitaliy, line 14: The word **Test** should be added at the end of the method name.

service-api/src/test/java/greencity/dto/placecomment/PlaceCommentAdminDtoTest.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class PlaceCommentAdminDtoTest {
+
+    @Test
+    void getSortableFields() {
```
ChernenkoVitaliy, line 14: Please rename it according to project test convention.

service-api/src/test/java/greencity/dto/ratingstatistics/RatingPointsDtoTest.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class RatingPointsDtoTest {
+
+    @Test
+    void getSortableFields() {
```
ChernenkoVitaliy, line 14: The word **Test** must be added at the end of the method name.

service-api/src/test/java/greencity/dto/ratingstatistics/RatingStatisticsDtoForTablesTest.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class RatingStatisticsDtoForTablesTest {
+
+    @Test
+    void getSortableFields() {
```
ChernenkoVitaliy, line 14: The same here.

core/src/main/java/greencity/controller/HabitCommentController.java
```diff
@@ -1,10 +1,6 @@
 package greencity.controller;
 
-import greencity.annotations.ApiPageable;
-import greencity.annotations.ApiPageableWithoutSort;
-import greencity.annotations.CurrentUser;
-import greencity.annotations.ImageArrayValidation;
-import greencity.annotations.ValidLanguage;
+import greencity.annotations.*;
```
ChernenkoVitaliy, line 3: You are not supposed to use wildcard.
Please, replace by usual imports.

core/src/main/java/greencity/validator/SortPageableValidator.java
```diff
@@ -0,0 +1,56 @@
...
+
+        if (validFields == null) {
+            throw new IllegalArgumentException(ErrorMessage.INVALID_DTO_CLASS);
+        }
+
+        sort.forEach(order -> {
+            String property = order.getProperty();
+            if (!validFields.contains(property)) {
+                throw new UnsupportedSortException(ErrorMessage.INVALID_SORTING_VALUE);
```
ChernenkoVitaliy, line 52: It would be great if You also added to the error message the cause of the error.

I mean, do something like this: `ErrorMessage.INVALID_SORTING_VALUE + property`

core/src/main/java/greencity/handler/CustomPageableHandlerMethodArgumentResolver.java
KostashchukIryna: You are right, forgot to change it here

core/src/main/java/greencity/handler/CustomSortHandlerMethodArgumentResolver.java
KostashchukIryna: Tried to follow principle "composition over inheritance". In addition, in the test there was problem with 
`super.resolveArgument(parameter, mavContainer, webRequest, binderFactory);`
so an easier way was to mock that method

core/src/main/java/greencity/validator/SortPageableValidator.java
KostashchukIryna: I agree with you that it might not be obvious. It's a part of a reason why I also added new interface Sortable and left the task of defining sortable fields to each dto itself. 
Using reflection was my first option but thought it was not best practice. I should probably ask if i can use it

core/src/test/java/greencity/validator/SortPageableValidatorTest.java
KostashchukIryna: You are right, didn't noticed i violated convention

service-api/src/test/java/greencity/dto/ratingstatistics/RatingStatisticsDtoForTablesTest.java
KostashchukIryna: You are right. I generated this test classes and forgot to modify the name. Thank you for pointing out

core/src/main/java/greencity/controller/HabitCommentController.java
KostashchukIryna: You are right, didn't noticed

core/src/main/java/greencity/validator/SortPageableValidator.java
KostashchukIryna: Okay, will do it

PR #8164, opened 2025-02-19
Fixed sort parameter validation

KostashchukIryna:
## Issue Link 📋
#8047
### Issue summary
Previously, if an invalid sort parameter was provided, it would result in a SQL error, which would trigger a 500 status code _or_ service would throw UnsupportedSortException if any field was passed in sort (both valid and invalid) even given that @ApiPageable allows sort.
Знімок екрана 2025-01-14 141911

### Solution
This PR proposes the validation of sorting parameters in the handler layer.

## Summary by CodeRabbit

- **New Features**
  - Improved API documentation for paginated responses by explicitly indicating the expected response structure.
  - Enhanced sorting functionality with added validation to ensure only permitted sort fields are used.

- **Refactor**
  - Consolidated pagination and sorting configurations across endpoints for more consistent API behavior.
  - Streamlined request processing by updating custom resolver logic.

- **Tests**
  - Introduced new tests for pagination and sorting validation while removing outdated sort exception tests.


core/src/main/java/greencity/config/PageableConfig.java
```diff
@@ -1,21 +1,50 @@
...
+    }
+
     /**
-     * Sets max page size for pageable objects.
+     * Adds custom pageable and sort resolvers to the list of argument resolvers.
      */
     @Override
     public void addArgumentResolvers(List<HandlerMethodArgumentResolver> argumentResolvers) {
-        PageableHandlerMethodArgumentResolver resolver = new PageableHandlerMethodArgumentResolver();
-        resolver.setMaxPageSize(100);
```
ChernenkoVitaliy, line 17: Is this restriction of `maxPageSize` not necessary anymore?

core/src/main/java/greencity/handler/CustomPageableHandlerMethodArgumentResolver.java
```diff
@@ -16,6 +17,12 @@
 import static greencity.constant.PageableConstants.SIZE;
 
 public class CustomPageableHandlerMethodArgumentResolver extends PageableHandlerMethodArgumentResolver {
+    private final CustomSortHandlerMethodArgumentResolver customSortResolver;
+
+    public CustomPageableHandlerMethodArgumentResolver(CustomSortHandlerMethodArgumentResolver customSortResolver) {
```
ChernenkoVitaliy: It is not necessary but You can use` @RequiredArgsConstructor` from `Lombok` instead of creating the constructor manually.

The same for `PageableConfig` and `CustomSortHandlerMethodArgumentResolver` classes.

core/src/main/java/greencity/handler/CustomSortHandlerMethodArgumentResolver.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public boolean supportsParameter(MethodParameter parameter) {
+        return Sort.class.equals(parameter.getParameterType());
+    }
+
+    @NonNull
+    @Override
+    public Sort resolveArgument(@NonNull MethodParameter parameter, ModelAndViewContainer mavContainer,
+        @NonNull NativeWebRequest webRequest, WebDataBinderFactory binderFactory) {
```
ChernenkoVitaliy: I don't know if it is important, but in the interface where this method was declared use `@Nullable`  annotation for method arguments and You are using `@NonNull` instead.

Also in the parent interface marked different fields from yours. You can look at the parent declaration and find differences.

core/src/main/java/greencity/annotations/SortableField.java
```diff
@@ -0,0 +1,22 @@
...
+ * handling pageable API requests.
+ * </p>
+ * <p>
+ * The validation mechanism ensures that only fields with this annotation are
+ * allowed in sorting parameters.
+ * </p>
+ **/
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SortableField {
```
ChernenkoVitaliy: Why do You need two identical annotations?
You have the same in the `service-api `module. You can use it in the `core` module easily.

core/src/main/java/greencity/handler/CustomPageableHandlerMethodArgumentResolver.java
KostashchukIryna: You are right, will remove boilerplate code

core/src/main/java/greencity/config/PageableConfig.java
KostashchukIryna: Yes, it is now handled in `CustomPageableHandlerMethodArgumentResolver`

ChernenkoVitaliy (approved):

maks741 (approved):

PR #8173, opened 2025-02-24
Fixed checkstyle for PlaceServiceImpl

ChernenkoVitaliy:
Fixer checkstyle deploy issue for PlaceServiceImpl.

## Summary by CodeRabbit

- **Refactor**
  - Improved internal code consistency and maintainability, ensuring the application’s core functionality remains reliable without impacting user experience.

### DIFF
--- a/service/src/main/java/greencity/service/PlaceServiceImpl.java
+++ b/service/src/main/java/greencity/service/PlaceServiceImpl.java
@@ -585,7 +585,7 @@ public class PlaceServiceImpl implements PlaceService {
      */
     @Override
     public PlaceResponse addPlaceFromUi(AddPlaceDto dto, String email) {
-        PlaceResponse placeResponse = modelMapper.map(dto, PlaceResponse.class);
+        final PlaceResponse placeResponse = modelMapper.map(dto, PlaceResponse.class);
         User user = userRepo.findByEmail(email)
             .orElseThrow(() -> new NotFoundException("User with email " + email + " doesn't exist"));
         if (user.getUserStatus().equals(UserStatus.BLOCKED)) {


service/src/main/java/greencity/service/PlaceServiceImpl.java
```diff
@@ -585,7 +585,7 @@ public List<FilterPlaceCategory> getAllPlaceCategories() {
      */
     @Override
     public PlaceResponse addPlaceFromUi(AddPlaceDto dto, String email) {
-        PlaceResponse placeResponse = modelMapper.map(dto, PlaceResponse.class);
+        final PlaceResponse placeResponse = modelMapper.map(dto, PlaceResponse.class);
```
Cr1stal423, line 588: Will it work fine if the variable is final? Below in this code block, we want to set it to the value

dxrknesss: I don't see where we change the reference for this variable.
image

dxrknesss: > Will it work fine if the variable is final? Below in this code block, we want to set it to the value

I don't see where we change reference for this variable.
image

That said, it should stay final. (checkstyle probably knows better than us :D)

dxrknesss (approved):

service/src/main/java/greencity/service/PlaceServiceImpl.java
ChernenkoVitaliy: A variable declared as final means that its value cannot be changed once it’s assigned.
I do not see where we try to assign a new value for it.

Please clarify the issue. Could you point to the line where can be a problem?
I will try to fix it.

Cr1stal423: Here we declare a variable: 
final PlaceResponse placeResponse = modelMapper.map(dto, PlaceResponse.class) and here we want to populate it with a value: placeResponse.setLocationAddressAndGeoDto(initializeGeoCodingResults(geocodingResults)); and locationAddressAndGeoDto is a field of the PlaceResponse class, won't there be a problem?

Cr1stal423: I did not switch to the branch locally, if the compiler pass it, then everything is fine

dxrknesss: > Here we declare a variable: final PlaceResponse placeResponse = modelMapper.map(dto, PlaceResponse.class) and here we want to populate it with a value: placeResponse.setLocationAddressAndGeoDto(initializeGeoCodingResults(geocodingResults)); and locationAddressAndGeoDto is a field of the PlaceResponse class, won't there be a problem?

if the field is not final, it's value could be set anytime.

variable placeResponse is final, which means we can't reference another object with it (cannot assign another object with = operator), but we can change fields freely.

ChernenkoVitaliy: No. There aren't issues with it.
We can easily set values for fields.

If it is final we **can't do the following**:
placeResponse = new PlaceResponse(); == here will be an error. But set fields we can.

There isn't any problem with compilation.

Cr1stal423: If PlaceResponse allows mutation, does it make sense to put final here? What does the checkstyle say about this? Because it only prohibits changing the reference, but not changing the state of the object.

Cr1stal423 (approved):

service/src/main/java/greencity/service/PlaceServiceImpl.java
ChernenkoVitaliy: Here is what checkstyle says:
checkstyle_issue
Here text of the issue if there is problem with picture:
"**, but allowed 3.  Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value).**"

I do not see any reason to make it final(maybe I don't see whole picture of problem) but checkstyle does.
If You have any idea how to fix it another way please let me know and I'ii do it.

Issue #8103, opened 2025-02-03
Implement Functionality for Retrieving Database Metadata and Data with Export Options.


ChernenkoVitaliy:
### Summary:
The goal of this task is to implement the functionality for the following:
1. **Retrieve metadata about database tables** (table names and columns).
2. **Retrieve rows from a specific table** by specifying table name, limit, and offset.
3. **Download data as an Excel file** based on the table name, limit, and offset.
4. Ensure proper security and validation of request parameters (e.g., `secretKey`).

### Tasks:
- **Create an endpoint to retrieve all database tables metadata:**
  - Implement a method to return a list of all tables in the database, along with their columns.

- **Create an endpoint to retrieve rows from a database table with limit and offset:**
  - Implement a method to retrieve rows from a specified table, with configurable `limit` and `offset` parameters.
  - The method should validate the table name using a regular expression (`^[A-Za-z_]+$`) and handle invalid inputs gracefully.

- **Create an endpoint to download data as an Excel file:**
  - Implement an endpoint that allows users to download rows from a database table as an Excel file.
  - The file should include data from the specified table, limited by the `limit` and `offset` parameters.
  - The filename should be dynamically generated based on the table name and the row range (e.g., `tableName(offset-limit).xlsx`).

- **Security & Validation:**
  - All endpoints should accept the `secretKey` in the request header for authorization(security with .env file).
  - Also should be secured by Spring Security. Only authorized users(any role) have access(+secretKey).
  - Ensure proper validation for request parameters such as `tableName`, `limit`, and `offset`.
  - Max row limit is **10_000**;

### Suggested Endpoints:
1. **`GET /export/settings/tables`**
   - Returns a list of all database tables and their columns.
   - Request Header: `secretKey=<value>`

2. **`GET /export/settings/select`**
   - Retrieves rows from the specified table with the given `limit` and `offset`.
   - Request Parameters: `tableName=<value>`, `limit=<int>`, `offset=<int>`
   - Request Header: `secretKey=<value>`

3. **`GET /export/settings/download-table-data`**
   - Downloads table data as an Excel file with the specified `tableName`, `limit`, and `offset`.
   - Request Parameters: `tableName=<value>`, `limit=<int>`, `offset=<int>`
   - Request Header: `secretKey=<value>`

PR #8148, opened 2025-02-13
8079 display log files

Warded120:
Issue link: [8079](https://github.com/ita-social-projects/GreenCity/issues/8079)

changed:
- created controller and service to manage logs
- created service for .env files
- added new appender in Logback configuration

## Summary by CodeRabbit

- **New Features**
  - Introduced secure log file management endpoints to list, view, download logs, and remove sensitive configuration in controlled environments.
  - Enhanced error responses by providing clearer, context-specific messages for log access and configuration issues.
  - Improved logging infrastructure with automatic file rotation and size management.
  - Added robust environment configuration management with secret key validation to ensure secure operations.
  - Added new constants for improved error messaging related to log files and configuration issues.
  - Introduced new data transfer objects for log file filtering and metadata handling.
  - Added new exception classes to handle specific error scenarios related to log file operations.

- **Chores**
  - Upgraded dependencies and configuration settings for improved environment variable management.

PR #8174, opened 2025-02-25
Feature #8103 - Implemented functionality for retrieving database metadata and data with export options.

ChernenkoVitaliy:
# GreenCity PR
## Issue Link 📋
https://github.com/ita-social-projects/GreenCity/issues/8103

## Added
- Added _org.apache.poi_ dependency in the service module;
- Added `TableRowsDto` and `TablesMetadataDto`;
- Added `ExportSettingsRepo` and `ExportSettingsRepoImpl` for receiving data about and from db;
- Created unit tests `ExportSettingsRepoImplTest`;
- Created `ExportSettingsService` interface and `ExportSettingsServiceImpl` for manipulation and preparing metadata and raw data from db;
- Created `ExportToFileService` interface and `ExportToFileServiceImpl` for creating files for export from data;
- Created `DatabaseMetadataException`, `FileGenerationException`, `InvalidLimitException` for export metadata and data from db functionality;
- Added unit test `ExportSettingsServiceImplTest`;
- Added unit tests `ExportToFileServiceImplTest`;
- Added `ExportSettingsController` with endpoints for receiving all table names, selecting data by limit, offset, and table name, and endpoint for exporting data in the file;

## Changed

- Moved _apache.poi.version_ property to the main `pom.xml`;
- Added necessary methods for new exceptions in the `CustomExceptionHandler`;
- Added constant for SQL_ROW_LIMIT in the `AppConstant`;
- Added new error messages in the `ErrorMessage`;
- Added method for receiving `TableMetadataDto` and `TableRowsDto` in the `ModelUtils` classes;

## Summary Of Changes 🔥
 Created functionality for receiving info about DB such as table names with column names, selecting rows by some criteria(table name, offset, limit), and downloading Excel files by the same criterias. All endpoints are secured by Spring security and additional security with the secret key based on .env functionality.

## Summary by CodeRabbit

- **New Features**
  - Launched a new export settings functionality that allows users to fetch table metadata, query specific data, and download the results as an Excel file.
  - Enhanced security measures with explicit access control for the export endpoints.
  - Improved error messaging to provide clearer feedback on invalid requests or export failures.

- **Tests**
  - Introduced comprehensive test suites to validate and ensure the robust operation of the new export features.

- **Chores**
  - Updated dependency versions to support improved file generation capabilities.
  - Added new constants and error messages for better handling of request parameters and exceptions.
  - Introduced new DTOs for managing table parameters and export data.


core/src/main/java/greencity/controller/ExportSettingsController.java
```diff
@@ -0,0 +1,99 @@
...
+        @ApiResponse(responseCode = "400", description = HttpStatuses.BAD_REQUEST,
+            content = @Content(examples = @ExampleObject(HttpStatuses.BAD_REQUEST)))
+    })
+    @GetMapping("/select")
+    public ResponseEntity<TableRowsDto> getSelected(@RequestParam @Pattern(regexp = "^[A-Za-z_]+$") String tableName,
+        @RequestParam int limit,
+        @RequestParam int offset,
+        @RequestHeader String secretKey) {
+        return ResponseEntity.ok(exportSettingsService.selectFromTable(tableName, limit, offset, secretKey));
+    }
```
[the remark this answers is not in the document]
ChernenkoVitaliy: All necessary validations are present in the `ExportSettingsServiceImpl`.

service-api/src/main/java/greencity/service/ExportSettingsService.java
```diff
@@ -0,0 +1,37 @@
...
+     * name, limit, offset.
+     *
+     * @param tableName {@link String}
+     * @param limit     is limit of rows to return.
+     * @param offset    is offset for rows to return.
+     *
+     * @return {@link InputStream} InputStream with file.
+     */
+    InputStream getExcelFileAsResource(String tableName, int limit, int offset, String secretKey);
+}
```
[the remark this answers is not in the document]
ChernenkoVitaliy: Thanks. Fixed.

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+            throw new InvalidLimitException(String.format(ErrorMessage.EXCEED_LIMIT, limit));
+        }
+
+        Workbook workbook = new XSSFWorkbook();
+        Sheet sheet = workbook.createSheet(tableName);
+        TableRowsDto data = exportSettingsRepo.selectPortionFromTable(tableName, limit, offset);
+        createHeaderRow(workbook, sheet, data);
+        populateTableCells(workbook, sheet, data);
+        return convertWorkbookToInputStream(workbook);
+    }
```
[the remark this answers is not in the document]
ChernenkoVitaliy: Thanks for the advice. You were right. Fixed.

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        font.setBold(true);
+
+        Set<String> raw = data.getTableData().getFirst().keySet();
+        int cellIndex = 0;
+        for (String key : raw) {
+            Cell headerCell = header.createCell(cellIndex++);
+            headerCell.setCellValue(key);
+            headerCell.setCellStyle(headerStyle);
+        }
+    }
```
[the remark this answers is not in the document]
ChernenkoVitaliy: Thanks. You were right. Fixed.

core/src/test/java/greencity/controller/ExportSettingsControllerTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        mockMvc.perform(get(SETTINGS_CONTROLLER_LINK + "/select")
+            .param("tableName", NOT_EXISTS_TABLE_NAME)
+            .param("limit", String.valueOf(LIMIT))
+            .param("offset", String.valueOf(OFFSET))
+            .header("secretKey", SECRET_KEY)
+            .accept(MediaType.APPLICATION_JSON))
+            .andExpect(status().isInternalServerError()) // Expect 404 Not Found for non-existent table
+            .andReturn();
+    }
```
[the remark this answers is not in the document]
ChernenkoVitaliy: Thanks. Fixed.

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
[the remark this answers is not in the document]
ChernenkoVitaliy: Thanks. Fixed

core/src/test/java/greencity/controller/ExportSettingsControllerTest.java
```diff
@@ -0,0 +1,180 @@
...
+            .param("limit", String.valueOf(invalidLimit))
+            .param("offset", String.valueOf(OFFSET))
+            .header("secretKey", SECRET_KEY)
+            .accept(MediaType.APPLICATION_JSON))
+            .andExpect(status().isBadRequest())
+            .andReturn();
+    }
+
+    @Test
+    void downloadExcelWithValidParamsTest() throws Exception {
```
maks741, line 161: I believe /download-table-data endpoint should also have unit tests for invalid input params, similar to the /select endpoint

core/src/main/java/greencity/controller/ExportSettingsController.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    @Operation(summary = "Get table rows by params.")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = HttpStatuses.OK,
+            content = @Content(schema = @Schema(implementation = TableRowsDto.class))),
+        @ApiResponse(responseCode = "400", description = HttpStatuses.BAD_REQUEST,
+            content = @Content(examples = @ExampleObject(HttpStatuses.BAD_REQUEST))),
+        @ApiResponse(responseCode = "401", description = HttpStatuses.UNAUTHORIZED,
+            content = @Content(examples = @ExampleObject(HttpStatuses.UNAUTHORIZED))),
+        @ApiResponse(responseCode = "403", description = HttpStatuses.FORBIDDEN,
```
Cr1stal423, line 65: There can't be a 403 here. Because in SecurityConfig all endpoints in the /settings/export/** path are available to all roles, this applies to all endpoints of this controller

service-api/src/main/java/greencity/dto/exportsettings/TableRowsDto.java
```diff
@@ -0,0 +1,12 @@
+package greencity.dto.exportsettings;
+
+import lombok.Builder;
+import lombok.Data;
+import java.util.List;
+import java.util.Map;
+
+@Data
+@Builder
+public class TableRowsDto {
```
Cr1stal423: Can this class be a record?

service-api/src/main/java/greencity/dto/exportsettings/TablesMetadataDto.java
```diff
@@ -0,0 +1,12 @@
+package greencity.dto.exportsettings;
+
+import lombok.Builder;
+import lombok.Data;
+import java.util.List;
+import java.util.Map;
+
+@Data
+@Builder
+public class TablesMetadataDto {
```
Cr1stal423: and it's also

service/src/main/java/greencity/service/ExportSettingsServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public InputStream getExcelFileAsResource(String tableName, int limit, int offset, String secretKey) {
+        dotenvService.validateSecretKey(secretKey);
+        checkLimitAndOffset(limit, offset);
+
+        return exportToFileService.exportTableDataToExcel(tableName, limit, offset);
+    }
+
+    private void checkLimitAndOffset(int limit, int offset) {
```
Cr1stal423: This method can be optimized

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private void populateTableCells(Workbook workbook, Sheet sheet, TableRowsDto data) {
+        CellStyle style = workbook.createCellStyle();
+        style.setWrapText(true);
+
+        Row headerRow = sheet.getRow(0);
+        List<Map<String, String>> tableData = data.getTableData();
+
+        int rowIndex = 1; // Start after the header row
```
Cr1stal423: Are these comments necessary here?

service/src/test/java/greencity/service/ExportToFileServiceImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+        Method method = ExportToFileServiceImpl.class.getDeclaredMethod("convertWorkbookToInputStream", Workbook.class);
+        method.setAccessible(true);
+        Workbook spyWorkbook = spy(new XSSFWorkbook());
+        doThrow(new IOException("Some exception message")).when(spyWorkbook).write(any(ByteArrayOutputStream.class));
+
+        try {
+            method.invoke(exportToFileService, spyWorkbook);
+            fail("Expected FileGenerationException to be thrown");
+        } catch (InvocationTargetException e) {
+            // Unwrap the InvocationTargetException to get the actual cause
```
Cr1stal423: And here

core/pom.xml
```diff
@@ -24,7 +24,6 @@
         <springdoc.version>2.3.0</springdoc.version>
         <thymeleaf.spring6.version>3.1.2.RELEASE</thymeleaf.spring6.version>
         <thymeleaf.extras.version>3.0.4.RELEASE</thymeleaf.extras.version>
-        <apache.poi.version>5.2.5</apache.poi.version>
```
RomaMocherniuk, line 27: why do you remove this?
I see it is used [here](https://github.com/ita-social-projects/GreenCity/blob/dev/core/pom.xml#L229)

core/src/main/java/greencity/config/SecurityConfig.java
```diff
@@ -292,7 +293,8 @@ public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Excepti
                     FRIENDS,
                     NOTIFICATIONS,
                     HABIT_ASSIGN_ID + "/friends/habit-duration-info",
-                    "/ai/**")
+                    "/ai/**",
+                    SETTINGS_EXPORT)
```
RomaMocherniuk, line 297: Why do we need this endpoint for users? 
Shouldn't it be available only to admins and moderators?

dao/src/main/java/greencity/repository/impl/ExportSettingsRepoImpl.java
```diff
@@ -0,0 +1,86 @@
...
+    public TablesMetadataDto getTablesMetadata() {
+        Map<String, List<String>> tablesMetaDada = new HashMap<>();
+
+        try (Connection connection = dataSource.getConnection()) {
+            DatabaseMetaData metaData = connection.getMetaData();
+            // null for database, schema, and catalog means the method will return all
+            // tables in the current database.
+            ResultSet tables = metaData.getTables(null, null, null, new String[] {"TABLE"});
+            while (tables.next()) {
+                List<String> columnsNames = new LinkedList<>();
```
RomaMocherniuk: Is there any reason for using Linked List here?

dao/src/main/java/greencity/repository/impl/ExportSettingsRepoImpl.java
```diff
@@ -0,0 +1,86 @@
...
+                String tableName = tables.getString("TABLE_NAME");
```
RomaMocherniuk: please more all stings to the constants

service-api/src/main/java/greencity/constant/ErrorMessage.java
```diff
@@ -246,4 +246,10 @@ public class ErrorMessage {
     public static final String BAD_SECRET_KEY = "The given secret key is incorrect";
     public static final String CANNOT_DELETE_DOTENV = "Failed to delete .env file";
     public static final String FUNCTIONALITY_NOT_AVAILABLE = "Functionality is not available";
+    public static final String NEGATIVE_LIMIT = "Limit cannot be negative";
+    public static final String NEGATIVE_OFFSET = "Offset cannot be negative";
```
RomaMocherniuk, line 250: I see you use it for tests only

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private InputStream convertWorkbookToInputStream(Workbook workbook) {
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+            workbook.write(byteArrayOutputStream);
+            workbook.close();
+
+            return new ByteArrayInputStream(byteArrayOutputStream.toByteArray());
+        } catch (IOException e) {
+            log.error(e.getMessage(), e);
+            throw new FileGenerationException(ErrorMessage.GENERATION_EXCEL_FILE_ERROR, e);
```
RomaMocherniuk, line 96: please add test for this case

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Set;
+
+@Slf4j
+@Service
+public class ExportToFileServiceImpl implements ExportToFileService {
+    @Transactional(readOnly = true)
+    @Override
+    public InputStream exportTableDataToExcel(TableRowsDto data) {
+        if (data.getTableData().isEmpty()) {
+            throw new ResourceNotFoundException(String.format(ErrorMessage.EMPTY_TABLE, data.getTableData()));
```
RomaMocherniuk: please add test for this case

service/src/test/java/greencity/service/ExportToFileServiceImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+    void exportTableDataToExcelIfTableIsEmptyTest() {
+        TableRowsDto emptyRow = TableRowsDto.builder().tableData(new LinkedList<>()).build();
+
+        assertThrows(ResourceNotFoundException.class,
+            () -> exportToFileService.exportTableDataToExcel(emptyRow));
+    }
+
+    @Test
+    void testExceptionCatchingDuringCreatingFileTest() throws Exception {
+        Method method = ExportToFileServiceImpl.class.getDeclaredMethod("convertWorkbookToInputStream", Workbook.class);
```
RomaMocherniuk: please do not use reflection
Please test private methods using public API

service/src/test/java/greencity/service/ExportToFileServiceImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+        Workbook spyWorkbook = spy(new XSSFWorkbook());
+        doThrow(new IOException("Some exception message")).when(spyWorkbook).write(any(ByteArrayOutputStream.class));
+
+        try {
+            method.invoke(exportToFileService, spyWorkbook);
+            fail("Expected FileGenerationException to be thrown");
+        } catch (InvocationTargetException e) {
+            // Unwrap the InvocationTargetException to get the actual cause
+            Throwable cause = e.getCause();
+            assertInstanceOf(FileGenerationException.class, cause);
```
RomaMocherniuk: please use `assertThrows`

RomaMocherniuk (changes requested):
Left some comments

core/pom.xml
ChernenkoVitaliy: I didn't remove it but moved it to the main pom.xml located at the project's root.
The reason is we use apache.poi dependency in two different modules and each of them used a different version and it was the reason that we had two versions of dependency for the same lib in one project. Now we have only one which is defined in the root pom.xml and can be used in any module.

core/src/main/java/greencity/config/SecurityConfig.java
ChernenkoVitaliy: The reason is that this functionality can be used not only dev team but testers team as well. Unknown which role can have any member of this team use this functionality. To avoid unauthorized access to this functionality we have an additional security layer such as using .env file with secretKey. It means that to reach this functionality you should be authorized using Spring security and pass the correct secret key as well.

If you think these endpoints should be available only for admins and moderators let me know and I'll fix it.

core/src/main/java/greencity/controller/ExportSettingsController.java
ChernenkoVitaliy: Actually, it is possible. In the service wich, this controller is using(`ExportSettingsService`) one more service(`DotenvService`) which validates _secretKey_ passed as a header param. If the secret key is not valid exception `BadSecretKeyException` will be thrown and it will be the reason for the 403 response code.

core/src/test/java/greencity/controller/ExportSettingsControllerTest.java
ChernenkoVitaliy: Thanks. Added.

dao/src/main/java/greencity/repository/impl/ExportSettingsRepoImpl.java
ChernenkoVitaliy: There is no reason to use it here. Replaced with `ArrayList`.
Thanks.

dao/src/main/java/greencity/repository/impl/ExportSettingsRepoImpl.java
ChernenkoVitaliy: Thanks. Done.

Except the "%" in the line` ResultSet columns = metaData.getColumns(null, null, tableName, "%");`.
Should I move it also?

service-api/src/main/java/greencity/constant/ErrorMessage.java
ChernenkoVitaliy: Thank you for pointing.
I added an exception where it is supposed to be used.

service-api/src/main/java/greencity/dto/exportsettings/TableRowsDto.java
ChernenkoVitaliy: There is no reason not tot do it.
Thanks. Fixed.

service/src/main/java/greencity/service/ExportSettingsServiceImpl.java
ChernenkoVitaliy: Give me a hint of how, and I will do it.

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
ChernenkoVitaliy: Unit test already added for this case.
It name is _exportTableDataToExcelIfTableIsEmptyTest_.

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
ChernenkoVitaliy: No, they aren't.
Removed.

service/src/main/java/greencity/service/ExportToFileServiceImpl.java
ChernenkoVitaliy: Test already added.
It is named _testExceptionCatchingDuringCreatingFileTest_

service/src/test/java/greencity/service/ExportToFileServiceImplTest.java
ChernenkoVitaliy: Removed.

service/src/test/java/greencity/service/ExportToFileServiceImplTest.java
ChernenkoVitaliy: Please give me a hint of how to do it properly and I will do my best to fix it.
I have already tried all the approaches which I was able to find and none of it didn't help me with it.

service-api/src/main/java/greencity/dto/exportsettings/TablesMetadataDto.java
ChernenkoVitaliy: Good point. Fixed.

service/src/main/java/greencity/service/ExportSettingsServiceImpl.java
Cr1stal423: You have two ifs in this method checking the same condition if x < 0 and if y < 0, you can combine this into a single if(x < 0 || y <  0)

ChernenkoVitaliy: Then I have next:
```
private void checkLimitAndOffset(int limit, int offset) {
        if (limit < 0 || offset < 0) {
            if (limit< 0) {
                throw new IllegalArgumentException(ErrorMessage.NEGATIVE_LIMIT);
            }
                throw new IllegalArgumentException(ErrorMessage.NEGATIVE_OFFSET);
        }
        if (limit > AppConstant.SQL_ROW_LIMIT) {
            throw new InvalidLimitException(String.format(ErrorMessage.EXCEED_LIMIT, AppConstant.SQL_ROW_LIMIT));
        }
    }
```
Something like this?

ChernenkoVitaliy: Or this one?

```
private void checkLimitAndOffset(int limit, int offset) {
        if (limit < 0 || offset < 0) {
            throw new IllegalArgumentException(offset < 0 ? ErrorMessage.NEGATIVE_OFFSET : ErrorMessage.NEGATIVE_LIMIT);
        }
        if (limit > AppConstant.SQL_ROW_LIMIT) {
            throw new InvalidLimitException(String.format(ErrorMessage.EXCEED_LIMIT, AppConstant.SQL_ROW_LIMIT));
        }
    }
```

Cr1stal423: I think the option with a ternary operator is better. But for such validation, I would do it at the annotation level

core/src/main/java/greencity/controller/ExportSettingsController.java
Cr1stal423: In this case, wouldn't 400 respons be better, because this is not a valid secret key and not unauthorized access?

ChernenkoVitaliy: Maybe you are right.
But in my opinion, such kind of functionality should be fixed in the new PR cause functionality with .env file was implemented in the other PR(https://github.com/ita-social-projects/GreenCity/pull/8148), and the best way is to create a ticket "Change request" or another and fix it.

If you think it should be fixed in this PR let me know and I'll fix it.

service/src/main/java/greencity/service/ExportSettingsServiceImpl.java
ChernenkoVitaliy: Please clarify. 
Which and where annotation? In controller or create a dto and mark fields with annotation or another option?

Should I remove the method `checkLimitAndOffset` and use only the annotating approach or leave both?

service/src/test/java/greencity/service/ExportToFileServiceImplTest.java
maks741: maybe that implementation is better:

```
@Test
void testExceptionCatchingDuringCreatingFileTest() throws Exception {
    Workbook spyWorkbook = spy(new XSSFWorkbook());
    doThrow(new IOException("Some exception message")).when(spyWorkbook).write(any(ByteArrayOutputStream.class));

    assertThrows(
            FileGenerationException.class,
            () -> ReflectionTestUtils.invokeMethod(
                    exportToFileService,
                    "convertWorkbookToInputStream",
                    spyWorkbook
            )
    );
}
```

it still uses reflection but at least it is ReflectionTestUtils class from a dedicated org.springframework.test.util package

core/src/main/java/greencity/controller/ExportSettingsController.java
Warded120: > In this case, wouldn't 400 respons be better, because this is not a valid secret key and not unauthorized access?

400 (BadRequest) won't be better in this case. maybe you meant 401 (Unauthorized), but it is still not good, because it is connected to permissions to access this endpoints via secretKey. secretkey is not connected to authorization

core/src/main/java/greencity/controller/ExportSettingsController.java
```diff
@@ -0,0 +1,107 @@
...
+            content = @Content(examples = @ExampleObject(HttpStatuses.FORBIDDEN)))
+    })
+    @GetMapping("/select")
+    public ResponseEntity<TableRowsDto> getSelected(@RequestParam @Pattern(regexp = "^[A-Za-z_]+$") String tableName,
```
Warded120: maybe change method name from **getSelected** to **selectFromTable**, similar to service method?

core/src/main/java/greencity/controller/ExportSettingsController.java
```diff
@@ -0,0 +1,107 @@
...
+            content = @Content(schema = @Schema(implementation = TableRowsDto.class))),
+        @ApiResponse(responseCode = "400", description = HttpStatuses.BAD_REQUEST,
+            content = @Content(examples = @ExampleObject(HttpStatuses.BAD_REQUEST))),
+        @ApiResponse(responseCode = "401", description = HttpStatuses.UNAUTHORIZED,
+            content = @Content(examples = @ExampleObject(HttpStatuses.UNAUTHORIZED))),
+        @ApiResponse(responseCode = "403", description = HttpStatuses.FORBIDDEN,
+            content = @Content(examples = @ExampleObject(HttpStatuses.FORBIDDEN))),
+    })
+    @GetMapping("/download-table-data")
+    public ResponseEntity<InputStreamResource> downloadExcel(
```
Warded120: same here, method name is not very informative. Maybe change it to be something like: **exportTableRowsAsExcel()** or similar? (my method name variant is still not very good, but maybe you have better ones)

service/src/main/java/greencity/service/ExportSettingsServiceImpl.java
Warded120: If your parameters (limit and offset) are primitives, then you can annotate the parameter in a controller as in the example:
```
@Validated // Required to enable validation for method parameters
@RestController
@RequestMapping("/accounts")
public class AccountController {

    @GetMapping
    public ResponseEntity<String> someEndpoint(
            @RequestParam @Min(value = 1, message = "Page number must be at least 1") int limit) {
        return ResponseEntity.ok(someFunc());
    }
}
```

You already used similar validation on controler parameter tableName:
`@RequestParam @Pattern(regexp = "^[A-Za-z_]+$") String tableName`

With this approach, you may remove some of the exceptions you created

core/src/main/java/greencity/controller/ExportSettingsController.java
```diff
@@ -0,0 +1,107 @@
...
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.RequestHeader;
+
+@RequiredArgsConstructor
```
Warded120, line 26: I am not sure but is it required to Annotate the controller with @Validated to enable validations on method parameters?
I mean his ones:
`@RequestParam @Pattern(regexp = "^[A-Za-z_]+$") String tableName,`

Warded120: Did you test your endpoints with invalid tableName parameters (manually or with JUnit)? is `@Pattern(regexp = "^[A-Za-z_]+$")` works?

ChernenkoVitaliy: Yes. I did.
Actually, it worked without annotation but returned only the 400 status bad request.
But after I've added the `@Validated` annotation it returns the 400 bad request plus an informative message.

However, I have changed my approach. Instead of adding annotation with validation in the controller, I've added `TableParamsRequestDto` and all fields marked with validation annotations inside dto. Annotation `@Validation` was added in the controller as well.

core/src/main/java/greencity/controller/ExportSettingsController.java
ChernenkoVitaliy: Thanks. Fixed

core/src/main/java/greencity/controller/ExportSettingsController.java
ChernenkoVitaliy: I like your variant.
Thanks. Fixed.

service/src/main/java/greencity/service/ExportSettingsServiceImpl.java
ChernenkoVitaliy: Thank you both for pointing.
I've added checks in the controller layer and removed `checkLimitAndOffset` from this service.
Also removed `InvalidLimitException` which was used in this method.
Extra tests for this method also were removed.

service/src/test/java/greencity/service/ExportToFileServiceImplTest.java
ChernenkoVitaliy: Test refactored.
Reflection removed.

service/src/test/java/greencity/service/ExportToFileServiceImplTest.java
ChernenkoVitaliy: Done.

service/src/main/java/greencity/service/ExportSettingsServiceImpl.java
ChernenkoVitaliy: **Main changes.** Added `TableParamsRequestDto` with fields marked annotations for validation.
Now this dto used in the controller and service. I find this approach the most reliable and convenient in this situation cause it will be validated in the controller and cause all fields marked with different validation annotations it will be passed in the service and we will be sure that all fields used in service are valid.

maks741 (approved):

RomaMocherniuk (approved):

Warded120 (approved):

oleksandra-bulhakova (approved):

PR #8178, opened 2025-02-26
Fixed bug when dislikes didnt assign to a EventDto response entity.

Andrii-Danylenko:
Fixed a bug where dislikes were not assigned to an event when retrieving it by ID.

## Summary by CodeRabbit

- **New Features**
  - Event details now include a like counter that displays the total number of positive reactions from users, enhancing the transparency and informative nature of event engagement. This update improves the user experience by providing clearer insights into community sentiment regarding each event.

### DIFF
--- a/service/src/main/java/greencity/service/EventServiceImpl.java
+++ b/service/src/main/java/greencity/service/EventServiceImpl.java
@@ -833,14 +833,11 @@ public class EventServiceImpl implements EventService {
         setFollowers(List.of(eventDto), userId);
         setSubscribes(List.of(eventDto), userId);
         eventDto.setCurrentUserGrade(currentUserGrade);
-        eventDto.setDislikes(event.getUsersDislikedEvents().size());
         return eventDto;
     }
 
     private EventDto buildEventDto(Event event) {
-        EventDto eventDto = modelMapper.map(event, EventDto.class);
-        eventDto.setDislikes(event.getUsersDislikedEvents().size());
-        return eventDto;
+        return modelMapper.map(event, EventDto.class);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


service/src/main/java/greencity/service/EventServiceImpl.java
```diff
@@ -833,11 +833,14 @@ private EventDto buildEventDto(Event event, Long userId) {
         setFollowers(List.of(eventDto), userId);
         setSubscribes(List.of(eventDto), userId);
         eventDto.setCurrentUserGrade(currentUserGrade);
+        eventDto.setDislikes(event.getUsersDislikedEvents().size());
         return eventDto;
     }
 
     private EventDto buildEventDto(Event event) {
-        return modelMapper.map(event, EventDto.class);
+        EventDto eventDto = modelMapper.map(event, EventDto.class);
```
Cr1stal423: As a recommendation not to manually set the value after mapping, you can see why dislikes are not mapped, because the field for dislikes is available in both dto and entity

Cr1stal423 (approved):

service/src/main/java/greencity/service/EventServiceImpl.java
Cr1stal423: But your option also applicable

maks741 (approved):

service/src/main/java/greencity/service/EventServiceImpl.java
Andrii-Danylenko: > But your option also applicable

It was a workaround. Probably, I will introduce new mapper object to map DTO properly. Thank you for a comment

maks741 (approved):

PR #8187, opened 2025-02-26
Feature/v2 like dislike methods

Andrii-Danylenko:
Added new v2 methods: dislike-v2 and like-v2. These methods return a response object that can be used without going back to the database

## Summary by CodeRabbit

- **New Features**
  - Users can now express reactions on events via enhanced like/dislike interactions.
  - Event details now display up-to-date counts of likes and dislikes, offering clearer feedback on event popularity.
- **Bug Fixes**
  - Improved API response handling for like and dislike actions, ensuring accurate status codes are returned.
- **Documentation**
  - Updated API documentation to reflect new like/dislike methods and their functionalities.
- **Tests**
  - Added new tests for the like and dislike actions utilizing the new version 2 endpoints.
- **Chores**
  - Introduced a utility method for creating predefined EventDto objects for testing purposes.

### DIFF
--- a/core/src/main/java/greencity/controller/EventController.java
+++ b/core/src/main/java/greencity/controller/EventController.java
@@ -353,6 +353,46 @@ public class EventController {
     }
 
     /**
+     * Updated method to like/unlike Event.
+     *
+     * @author Andrii Danylenko
+     */
+    @Operation(summary = "Like/unlike Event and get response entity")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = HttpStatuses.OK),
+        @ApiResponse(responseCode = "400", description = HttpStatuses.BAD_REQUEST,
+            content = @Content(examples = @ExampleObject(HttpStatuses.BAD_REQUEST))),
+        @ApiResponse(responseCode = "401", description = HttpStatuses.UNAUTHORIZED,
+            content = @Content(examples = @ExampleObject(HttpStatuses.UNAUTHORIZED))),
+        @ApiResponse(responseCode = "404", description = HttpStatuses.NOT_FOUND,
+            content = @Content(examples = @ExampleObject(HttpStatuses.NOT_FOUND)))
+    })
+    @PostMapping("/{eventId}/like-v2")
+    public ResponseEntity<EventDto> likeV2(@PathVariable Long eventId,
+        @Parameter(hidden = true) @CurrentUser UserVO user) {
+        return ResponseEntity.ok().body(eventService.likeV2(eventId, user));
+    }
+
+    /**
+     * Updated method to dislike Event.
+     *
+     * @author Andrii Danylenko
+     */
+    @Operation(description = "Dislike event and get response entity")
+    @ApiResponses(value = {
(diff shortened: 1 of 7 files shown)


core/src/main/java/greencity/controller/EventController.java
```diff
@@ -352,6 +352,46 @@ public void like(@PathVariable Long eventId, @Parameter(hidden = true) @CurrentU
...
+     *
+     * @author Andrii Danylenko
+     */
+    @Operation(description = "Dislike event and get response entity")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = HttpStatuses.OK),
+        @ApiResponse(responseCode = "400", description = HttpStatuses.BAD_REQUEST,
+            content = @Content(examples = @ExampleObject(HttpStatuses.BAD_REQUEST))),
+        @ApiResponse(responseCode = "401", description = HttpStatuses.UNAUTHORIZED,
+            content = @Content(examples = @ExampleObject(HttpStatuses.UNAUTHORIZED)))
```
maks741: 404 response code should also be documented, similar to method above

core/src/test/java/greencity/controller/EventControllerTest.java
maks741, line 1: Is it possible to make dislikev2Test and likev2Test more precise? E.g. by instantiating mock EventDto instance, and ensuring that it is returned from the service via when(...) call. Then we can verify that we get expected EventDto instance as a result, instead of a simple null check

Andrii-Danylenko: Will do! Thank you for a review

Andrii-Danylenko: Due to the fact that this goes into the service level, I did these tests in EventServiceImplTest. Will commit soon

core/src/test/java/greencity/controller/EventControllerTest.java
```diff
@@ -392,6 +391,40 @@ void dislikeTest() throws Exception {
         verify(eventService).dislike(userVO, 1L);
     }
 
+    @Test
+    void dislikev2Test() throws Exception {
+        UserVO userVO = getUserVO();
+        EventDto eventDto = getEventDto();
+        ObjectMapper objectMapper = new ObjectMapper();
```
dxrknesss: I beg you, please refactor fixture creation to setUp() method😭

this object is created almost in every method of this test class, refactor it as field and create it in setUp()

dxrknesss: It should probably even be as static final field, since it doesn't have state.

https://stackoverflow.com/a/3909846

according to this SO answer, if you don't need writing capabilities, you can declare private static final ObjectReader field and be good!

the main point - refactor it's creation from each test

maks741 (approved):

dxrknesss (approved):

Cr1stal423:
I can't leave a comment on this code, but I think you should use SCREAMING_SNAKE_CASE for constants, and also remove the access modifier from the test of this class
image

Cr1stal423 (approved):

PR #8266, opened 2025-03-20
Feature/add a block with events

NazarVavrushchak:
## Summary by CodeRabbit

- **New Features**
  - The management page now displays two dedicated sections for "Organized Events" and "Attended Events," complete with event details such as creation date and status.
  - User interface labels have been updated and localized (including Ukrainian), enhancing clarity and ease of use when reviewing event-related information.
  
- **Bug Fixes**
  - Improved error handling with specific messages for null user IDs and cases where no events are found for a user.


DmytroLysenko1 (approved):

service/src/main/java/greencity/service/EventServiceImpl.java
```diff
@@ -1120,6 +1120,33 @@ public Page<UserProfilePictureDto> getUsersDislikedEventPage(Long eventId, Pagea
...
+        return organizedEvents.stream()
+            .map(event -> buildEventDto(event, userId))
+            .toList();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<EventDto> getAllEventsAttendedByUser(Long userId) {
```
RomaMocherniuk, line 1143: can we create a query to get all user events(organized and attended) by userId?
In this case, we do not need extra calls to DB, like `userRepo.findById(userId) `and `userRepo.findById(userId)`
Can we have a situation when a user is an organizer of some event but now attended of that event?
If not we can get all attended an  event by a user and then check if the user is an organizer of the event

softservedata (approved):

Andrii-Danylenko (changes requested):
Please, add some more tests to increase code coverage

Gubernatorr (approved):

softservedata (approved):

service/src/main/java/greencity/service/EventServiceImpl.java
NazarVavrushchak: Done

Andrii-Danylenko (approved):

PR #8501, opened 2025-06-10
Secure endpoinds for WebClient

oleksandra-bulhakova:
Secured endpoints created for the GreenCityUser microservice to use them in GreenCityRemoteClient

### DIFF
--- a/core/src/main/java/greencity/config/SecurityConfig.java
+++ b/core/src/main/java/greencity/config/SecurityConfig.java
@@ -224,8 +224,6 @@ public class SecurityConfig {
                     ACHIEVEMENTS,
                     ACHIEVEMENTS + COUNT,
                     ACHIEVEMENTS + CATEGORIES,
-                    ACHIEVEMENTS + "/all",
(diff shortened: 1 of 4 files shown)


core/src/main/java/greencity/config/SecurityConfig.java
```diff
@@ -224,8 +224,6 @@ public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Excepti
                     ACHIEVEMENTS,
                     ACHIEVEMENTS + COUNT,
                     ACHIEVEMENTS + CATEGORIES,
-                    ACHIEVEMENTS + "/all",
```
ChernenkoVitaliy, line 227: If you restrict the USER role to reach endpoints, it would be nice to add the necessary Swagger doc to the related controller. I mean not only for this endpoint but for all restricted in this PR.

oleksandra-bulhakova: Done

ChernenkoVitaliy (approved):

maks741 (approved):

